PR #2, opened 2016-09-29
Upgrade ruby version

ptrikutam:
Update the ruby version that is used on CircleCI.


ptrikutam:
SELF REVIEWING CODE LOOKS GOOD TO ME

PR #4, opened 2016-09-29
Update readme

ptrikutam:


gtourtellot:
README looks good, although there's no info about obtaining ruby.
E.g., "brew install ruby", or "brew upgrade ruby" if already installed.

For the foreman command to work, the foreman gem needs to be added to the Gemfile, right?

ptrikutam:
Yep -- there will be a `./bin/setup` file to follow which should do everything we need (install foreman, ruby, etc).

PR #17, opened 2016-10-27
Don't set a default username in the db

adarsh:
Reason for Change
=================
* Usernames are not required for postgres databases, particularly in development.
* Also: I never set a `root` user for for my development/test databases since it's not necessary.
* Also YAML doesn't support overriding when inheriting, because it's basically just merging hashes.

Changes
=======
* Remove the `root` database username from the default configuration.
* Add it to the `production` group.

Recommended Reviewers
=====================
@ptrikutam

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -21,8 +21,10 @@ default: &default
   # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: 5
   host: localhost
-  username: root
   database: commons
+
+production:
+  <<: *default
 
 development:
   <<: *default
@@ -34,8 +36,3 @@ development:
 test:
   <<: *default
   database: commonstest
-
-# TODO: Do we need to include this?
-# production:
-#   <<: *default
-#   database: db/production.sqlite3


config/database.yml
```diff
@@ -21,9 +21,12 @@ default: &default
   # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: 5
   host: localhost
-  username: root
   database: commons
 
+production:
+  <<: *default
+  username: root
```
ptrikutam: How will this work when we're on a Heroku DB? Will this override the username provided by `ENV['DATABASE_URL']`?

ptrikutam: Follow up Q: why is it necessary to include at all if there is an `ENV['DATABASE_URL']` present? On the chance that it isn't present on prod?

ptrikutam: To clarify, for my second q, I'm asking why the production details itself are necessary at all, not just the username.

adarsh: Yeah, I'm not sure that we need to specify any of this since Heroku uses the `ENV['DATABASE_URL']` value, which includes a username and pw:


> The DATABASE_URL for the Heroku Postgres add-on follows this naming convention:
> `postgres://<username>:<password>@<host>/<dbname>`
 
https://devcenter.heroku.com/articles/heroku-postgresql#spring-java

I've removed the username in fa94c7b. Let's see if it deploys!

PR #35, opened 2016-11-16
Add a bummr for more easily updating gems

adarsh:
Reason for Change
=================
* Gems should stay updated.
* Sometimes, they break the app.
* [`bummr`][1] helps us update the gems and figure out which ones break the app.

[1]: https://github.com/lpender/bummr

Changes
=======
* Add the gem
* Add a `.bummr-build.sh` file, which is how it knows how to run the tests.
* Describe the change in the `README.md`.

### DIFF
--- a/.bummr-build.sh
+++ b/.bummr-build.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+bundle exec rake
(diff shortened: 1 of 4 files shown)


ptrikutam:
I like the idea, though I'm a bit concerned about how much of an early stage this gem is in. Do you feel confident it'll be supported (or at least work well) throughout the duration of our project?

adarsh:
@ptrikutam Fair question. It seems to be under active development with a new release 8 days ago ([`git log origin/master`][1]) and at worst, if it fails, we can just go back to good ol' `bundle update`.

[1]: https://github.com/lpender/bummr/commits/master

Issue #33, opened 2016-11-07
Run elasticsearch callbacks asynchronously.


adarsh:
Originally discussed here: https://github.com/greencommons/commons/issues/25#issuecomment-258906665

If we set these callbacks to run asynchronously (e.g. spin up a Sidekiq job to change the search index), we can have a bit more control over when we actually update the index.

For example, we could handle errors more gracefully. Or just not execute the job in development or test environments unless some ENV is set.

https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#asynchronous-callbacks

PR #36, opened 2016-11-17
Run Search Index actions asynchronously

adarsh:
Reason for Change
=================
* When we perform CRUD actions on an AR model, we want to subsequently update the search index to keep everything in sync.
* That can delay the request cycle, so we should perform those activities in the background.

Changes
=======
* Add custom callbacks instead of the stock ElasticSearch ones.
* Send modified records to a `SearchIndex` service class to abstract away everything about ElasticSearch. Let it enqueue jobs or whatever it wants.
* Create a general job class to add/remove records from the search index. Built it to allow for any type of AR record with `ElasticSearch::Model` mixed in.
* Add `sidekiq` gem for background job processing.
* Add a way to disable the callbacks via `ENV`.

Minor
=====
* Add [`climate_control`] gem for temporarily modifying `ENV` in tests.
* Use [`pry-byebug`] instead of other pry/byebug variants.
* Add an RSpec tag to disable search indexing callbacks for tests.
* Add an RSpec tag to exercise background jobs inline.
* Allow Rubocop to have longer specs than 5 lines
* Use implicit FactoryGirl calls (e.g. no need for `FactoryGirl.` all the time)
* [Let RSpec infer the spec type based on the folder][1]. No need for `type: model`.

[1]: https://www.relishapp.com/rspec/rspec-rails/v/3-5/docs/directory-structure

[`climate_control`]: https://github.com/thoughtbot/climate_control
[`pry-byebug`]: https://github.com/deivid-rodriguez/pry-byebug

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,14 +12,13 @@ gem 'pg'
 gem 'puma', '~> 3.0'
 gem 'rails', '~> 5.0.0', '>= 5.0.0.1'
 gem 'sass-rails', '~> 5.0'
+gem 'sidekiq'
 gem 'uglifier', '>= 1.3.0'
 
 group :development do
   gem 'bummr'
   gem 'bundler-audit'
   gem 'listen', '~> 3.0.5'
-  gem 'pry-coolline'
-  gem 'pry-rails'
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
   gem 'web-console'
@@ -27,10 +26,11 @@ end
 
 group :development, :test do
   gem 'awesome_print'
-  gem 'byebug', platform: :mri
+  gem 'climate_control'
   gem 'database_cleaner'
   gem 'dotenv'
   gem 'factory_girl_rails'
+  gem 'pry-byebug'
   gem 'rspec-rails', '~> 3.5'
   gem 'rubocop'
   gem 'rubocop-rspec'
--- a/app/jobs/indexer_job.rb
+++ b/app/jobs/indexer_job.rb
@@ -1,0 +1,24 @@
+class IndexerJob
+  include Sidekiq::Worker
+  sidekiq_options queue: 'elasticsearch', retry: true
+
+  def perform(action, model_name, id)
+    @action = action
+    @id = id
+    @model_name = model_name
+
+    record.__elasticsearch__.send(index_action)
+  end
+
+  private
+
+  attr_reader :action, :id, :model_name
+
+  def record
+    model_name.constantize.find(id)
+  end
+
+  def index_action
+    "#{action}_document".to_sym
+  end
+end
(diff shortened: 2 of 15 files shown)


adarsh:
Addresses https://github.com/greencommons/commons/issues/33

app/service/search_index.rb
```diff
@@ -0,0 +1,17 @@
+class SearchIndex
+  def self.add(record)
+    if search_index_callbacks_enabled?
```
ptrikutam: Should there be some kind of error or alert if this isn't enabled? Error seems excessive, since we don't want a save to break because of indexing being disabled... It's OK if the answer is "nopes, we fail silently and get on with our lives." What do you think?

app/service/search_index.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def self.remove(record)
+    if search_index_callbacks_enabled?
+      IndexerJob.perform_async(:delete, record.id)
+    end
+  end
+
+  def self.search_index_callbacks_enabled?
+    ENV.fetch('ENABLE_SEARCH_INDEX_CALLBACKS', true) != 'false'
```
ptrikutam: Actually, on second thought, this probably doesn't need to be added to `.env.sample`. Since it defaults to true I think we're good.

Gemfile
```diff
@@ -12,25 +12,25 @@ gem 'pg'
 gem 'puma', '~> 3.0'
 gem 'rails', '~> 5.0.0', '>= 5.0.0.1'
 gem 'sass-rails', '~> 5.0'
+gem 'sidekiq'
```
adarsh: Yeah, I mean, a lot of the rest of it won't work without this gem.

app/jobs/indexer_job.rb
```diff
@@ -0,0 +1,24 @@
+class IndexerJob
+  include Sidekiq::Worker
+  sidekiq_options queue: 'elasticsearch', retry: false
```
adarsh: I guess we wouldn't be editing the records that much, so we should actually probably set `retry: true` to ensure they would be indexed.

The usual use case I'm guessing is a lot more updates so if we miss one, hey, no big deal.

Also, I was considering something like a catch-all regular nightly task to just re-index all the resources, regardless. Wouldn't be a huge deal and would ensure everything is up to date. Dunno, just spitballin'.

I reverted this to true in dd9fa6c.

app/service/search_index.rb
adarsh: How about adding a logger step?

The other option is the one I mentioned before, which is running a "re-index all the things every night at 3am" rake task, just in case.

Which do you think makes more sense?

adarsh: I added a logger step with a searchable tag, `[ELASTICSEARCH]` so we can find it easily.

ptrikutam: Logger sounds great. Also less "catastrophic failure"-y.

PR #41, opened 2016-11-30
Correct Sidekiq configuration

adarsh:
Reason for Change
=================
* Sidekiq jobs were not working correctly in development for a few reasons:
1. The interface for `IndexerJob` was not being used correctly.
1. There was no `worker` line in the `Procfiles`, so no workers were initiated. Hence, no jobs were being done.
1. Sidekiq was not initialized to use the correct redis connection URL.

Changes
=======
* Add configuration files and initializer for sidekiq for correct functioning.
* Add worker directives to foreman's `Procfile` manifest.
* Correct an interface error when creating `IndexerJob`s.

Minor
-----
* The "hey you've disabled the Elasticsearch callbacks" warning was inverted and incorrect. Fixed here.
* Add the sidekiq web monitoring page at `/sidekiq`, but only in development.

https://trello.com/c/f0lpURT0/27-add-a-search-field-and-return-a-result

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -C config/sidekiq.yml
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,4 @@
 web: bundle exec puma -C config/puma.rb
 worker: bundle exec sidekiq -C config/sidekiq.yml
 search: elasticsearch
+redis: redis-server
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -C config/sidekiq.yml
 search: elasticsearch
--- a/app/service/search_index.rb
+++ b/app/service/search_index.rb
@@ -1,23 +1,27 @@
 class SearchIndex
   def self.add(record)
     if search_index_callbacks_enabled?
-      IndexerJob.perform_async(:index, record.id)
+      IndexerJob.perform_async(:index, record.class.name, record.id)
     end
   end
 
   def self.remove(record)
     if search_index_callbacks_enabled?
-      IndexerJob.perform_async(:delete, record.id)
+      IndexerJob.perform_async(:delete, record.class.name, record.id)
     end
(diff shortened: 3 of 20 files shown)


Procfile
```diff
@@ -1 +1,2 @@
-web: bundle exec puma -C config/puma.rb
\ No newline at end of file
+web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -C config/sidekiq.yml
```
ptrikutam: We'll probably need to start paying for a worker on Heroku :(

app/service/search_index.rb
```diff
@@ -1,23 +1,27 @@
 class SearchIndex
   def self.add(record)
     if search_index_callbacks_enabled?
-      IndexerJob.perform_async(:index, record.id)
+      IndexerJob.perform_async(:index, record.class.name, record.id)
```
ptrikutam: Yeesh, can't believe I missed this. Good catch.

Procfile.dev
```diff
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -C config/sidekiq.yml
```
ptrikutam: Can we also add a process for redis?

```
redis: redis-server
```

ptrikutam: We may also want to add `dump.rdb` to the `.gitignore` when we do this.

adarsh:
> EDIT: I checked out a different branch and restarted foreman, and jobs started firing. I then ran into this issue (after previously manually deleting record 100): https://cl.ly/2x2c1r1I0W26

So that pic is showing errors where the job is failing because it's trying to find record 100 but it doesn't exist. That's a bug for us/me - I'm asking ES to remove a record from the index which does not exist in the database anymore :(

I'm opening an issue for this. Solution is something like `ResourceDeleter.remove` which will first remove the record from the index, *then* delete the record.

Good catch! Made an issue here: https://github.com/greencommons/commons/issues/43

Procfile.dev
adarsh: Yeah good call. Done in dd48e3b

adarsh:
@ptrikutam note that the other branch is a PR into this branch, so I will **merge it second**.

PR #42, opened 2016-11-30
Add Simple Search Results

adarsh:
Reason for Change
=================
* We want to set up basic search functionality across Rails/Elasticsearch.
* Also, we'd like to establish one happy-path integration test to ensure the main functionality of the page is working.

Changes
=======
* Add documentation for setting up QT and XCode.
* Add capybara-webkit gem for a headless web driver to simulate rendering a full page.

Minor
=====
* Remove whitespace.
* Add `slim` gem for alternative templating.
* Delete placeholder static page.
* Reduce concurrency to keep up with database connection pool.
* Add logging for debugging
* Allow for up to 3 expectations in one test block.
* Remove duplicate `rspec-rubocop` configuration.

https://trello.com/c/f0lpURT0/27-add-a-search-field-and-return-a-result

image

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,0 +1,5 @@
+class SearchController < ApplicationController
+  def new
+    @results = Resource.search(params[:query])
+  end
+end
--- a/app/jobs/indexer_job.rb
+++ b/app/jobs/indexer_job.rb
@@ -7,6 +7,12 @@ class IndexerJob
     @id = id
     @model_name = model_name
 
+    Rails.logger.tagged('ELASTICSEARCH') do
+      Rails.logger.warn(
+        "Hey! Performing a #{action} on a #{model_name} record with id# #{record.id}"
+      )
+    end
+
     record.__elasticsearch__.send(index_action)
   end
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,5 +50,7 @@ RSpec/ExampleLength:
   Description: Checks for long examples.
   Enabled: true
   Max: 20
-RSpec/FilePath:
-  Enabled: false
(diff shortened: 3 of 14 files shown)


app/controllers/search_controller.rb
```diff
@@ -0,0 +1,5 @@
+class SearchController < ApplicationController
+  def new
+    @results = Resource.search(params[:query])
```
ptrikutam: Do we not use strong parameters in the controller here?

app/jobs/indexer_job.rb
```diff
@@ -7,6 +7,12 @@ def perform(action, model_name, id)
     @id = id
     @model_name = model_name
 
+    Rails.logger.tagged('ELASTICSEARCH') do
+      Rails.logger.warn(
+        "Hey! Performing a #{action} on record number #{record.id}"
```
ptrikutam: Might be worth adding in the model name to this message too.

app/controllers/search_controller.rb
adarsh: Strictly speaking, we only need SP's for update/create/delete actions, but good eye.

adarsh:
> I'm sure I'm doing something wrong, but I'm getting the following error when trying to search: https://cl.ly/131G2h3A0Y1S

No this is a bug. The column is `null: true` so the metadata could be empty.

I'm going to add a migration to make it `null: false` and default to an empty hash.

app/jobs/indexer_job.rb
adarsh: Yes! Done in aa927fe

adarsh:
Changed the metadata column in 8a0ba63, please review the migration.

PR #51, opened 2016-12-02
Add elasticsearch resource index deletion task

ptrikutam:
# Reason for change

There are many times we want to clear the index for ElasticSearch & start over. We want to create a utility task for this.

# Changes

- add `elasticsearch.rake` with the task to delete the "resources" index

### DIFF
--- a/lib/tasks/elasticsearch.rake
+++ b/lib/tasks/elasticsearch.rake
@@ -1,9 +1,11 @@
 namespace :elasticsearch do
-  desc 'Delete the "resource" index'
-  task :delete_resource_index do
+  desc 'Deletes the "resource" index and regenerates it with all records currently in Resource'
+  task :reset_resource_index do
     client = Elasticsearch::Client.new(
       url: ENV.fetch('BONSAI_URL', 'http://localhost:9200'), log:true
     )
     client.indices.delete index: 'resources'
+    Resource.__elasticsearch__.create_index!
+    Resource.import
   end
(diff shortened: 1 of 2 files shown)


lib/tasks/elasticsearch.rake
```diff
@@ -0,0 +1,9 @@
+namespace :elasticsearch do
+  desc 'Delete the "resource" index'
+  task :delete_resource_index do
+    client = Elasticsearch::Client.new(
```
adarsh: Is there a way to access the existing client instance without creating a new one?

Heroku will run take tasks in their own dynos so it doesn't matter, I'm more curious if it's possible.

ptrikutam: @adarsh Good question.

The only place I see it being instantiated is in `config/initializers/elasticsearch.rb`, where we assign it to `Elasticsearch::Model.client`. 

It feels incorrect to access that in order to access ES. What do you think?

adarsh: I'm not sure I see it as correct or incorrect, but I can't explain why. I'm sure it's fine either way.

I was more just curious if you can use the same client instance for simplicity but there's really no need.

ptrikutam:
@adarsh made some more changes to ensure that the index always gets recreated when we try to reset it. Can you take a look and provide thoughts?

adarsh:
I love it.

PR #55, opened 2016-12-02
Add styles to search page

ptrikutam:
# Reason for change

The search page could use a little visual enhancing.

# Changes

- Update `SearchController` to not search the Resource index unless there is a query parameter present.
- Update the search field & search results page to be a bit more bootstrap-y. These are by no means final styles but might look a little nicer for proof-of-concepts.

# Screenshots
- Before: https://cl.ly/1l44212n1n2C
- After: https://cl.ly/3e3F3K3y2p0H

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,5 +1,5 @@
 class SearchController < ApplicationController
   def new
-    @results = Resource.search(params[:query])
+    @results = params[:query].present? ? Resource.search(params[:query]) : []
   end
 end
--- a/app/views/search/new.html.slim
+++ b/app/views/search/new.html.slim
@@ -1,21 +1,31 @@
-h1.title
-  | Search#new
+.row
+  .col-sm-12.col-md-10.col-md-1.col-lg-8.col-lg-offset-2.text-center
+    = form_for new_search_path, method: :get, html: { class: 'form-horizontal customer-search-form' } do |form|
+      .form-group
+        .col-xs-8
+          = text_field_tag :query, params[:query], placeholder: 'Enter Search Term', class: 'form-control input-lg'
+          small
+            em Try searching for "wind," "green," or "river."
+        .col-xs-4
+          = submit_tag 'Search', name: nil, class: 'col-lg-12 btn btn-lg btn-primary'
 
-.customer-search-form
-  = form_for new_search_path, method: :get do |form|
-    = form.label 'Search for'
-    p
-    = text_field_tag :query, params[:query]
-    = submit_tag 'Go', name: nil
 
-ul.search-results
+- unless @results.empty?
   - @results.each do |result|
-    h2.Result
-
-    li.search-result
-      h3.title
-        = result.title
-      h5.author-name
-        = result.metadata.fetch(:author, 'No Author Found')
-      h5.year-published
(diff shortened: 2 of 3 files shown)


app/controllers/search_controller.rb
```diff
@@ -1,5 +1,5 @@
 class SearchController < ApplicationController
   def new
-    @results = Resource.search(params[:query])
+    @results = params[:query].present? ? Resource.search(params[:query]) : []
```
adarsh: Minor: Can we make this not a ternary :) I find it easier to read and also it exposes costly conditional branching logic.

app/views/search/new.html.slim
```diff
@@ -1,21 +1,29 @@
...
+          dd
+            = result.id
+          dt
+            strong Author
+          dd
+            = result.metadata.fetch(:author, 'No Author Found')
+          dt
+            strong Published
+          dd
+            = result.metadata.fetch(:year_published, 'Year Published Unknown')
```
adarsh: Would you mind adding some helpful Before/After screencaps? 🖌

app/views/search/new.html.slim
```diff
@@ -1,21 +1,29 @@
...
-      h3.title
-        = result.title
-      h5.author-name
-        = result.metadata.fetch(:author, 'No Author Found')
-      h5.year-published
-        = result.metadata.fetch(:year_published, 'Year Published Unknown')
+    .row
+      .well.col-sm-12
+        h3.text-center
+          = "#{result.title}"
```
adarsh: I'm probably missing something - why does `result.title` need interpolation?

app/views/search/new.html.slim
```diff
@@ -1,21 +1,29 @@
...
 
-.customer-search-form
-  = form_for new_search_path, method: :get do |form|
-    = form.label 'Search for'
-    p
-    = text_field_tag :query, params[:query]
-    = submit_tag 'Go', name: nil
 
-ul.search-results
+- unless @results.empty?
```
adarsh: Not now, but we should return a `NullResult` object at some point, which is just an empty array. 

https://robots.thoughtbot.com/rails-refactoring-example-introduce-null-object

app/views/search/new.html.slim
ptrikutam: Ah -- remnant from a piece I'd worked on and then abandoned. Will fix.

app/controllers/search_controller.rb
ptrikutam: Would this be acceptable?

```
@results = if params[:query].present?
             Resource.search(params[:query])
           else
             []
           end
```

adarsh: The alternative is to more declaratively just call `Resource.search(params[:query])` and then over in the model, something like

```ruby
def self.search(query)
  if query
    super(query)
  else
    NullQuery.new.search # => []
  end
end

adarsh: But these are minor things, this should merge regardless. Null objects can come later.

ptrikutam: Got it. OK, yeah I suspected that it would need to be something existing outside of the controller.

app/views/search/new.html.slim
ptrikutam: Noted, thanks.

PR #56, opened 2016-12-07
Initial ETL work

adarsh:
Reason for Change
=================
* External `.epub` files need to be imported into our database.
* The `kiba` framework helps us define source, transformation, and destination classes for modular re-use.
* This PR is a proof-of-concept for importing two local `.epub` files.

Changes
=======
* Add `kiba` and `epub-parser` gems.
* Add a new `content` JSONB column to the `Resources` model. Putting it all into `metadata` felt cramped. Also when testing output, the content volume rendered terminal-based debugging crazy-making.
* Add a kiba file which defines the order of operations and sources the `common.rb` manifest file of helper classes.

Minor
-----
* Add documentation to the `README.md` and new `ETL.md` documentation files.

https://trello.com/c/468blsIv/21-1-1-1-1-document-ingestion-indexing-est-8d


db/migrate/20161207220901_add_content_to_resources.rb
```diff
@@ -0,0 +1,5 @@
+class AddContentToResources < ActiveRecord::Migration[5.0]
+  def change
+    add_column :resources, :content, :jsonb, null: false, default: {}
```
ptrikutam: Out of curiosity, what is the motivation for using a `:jsonb` field vs. a large text field?

lib/etl/page_content_extractor.rb
```diff
@@ -0,0 +1,21 @@
...
+  def initialize(parsed_book)
+    @book_text = ''
+    @parsed_book = parsed_book
+  end
+
+  def start
+    parsed_book.each_page_on_spine do |page|
+      print '.'
+      parsed_xml = page.content_document.read
+      page_plain = ActionView::Base.full_sanitizer.sanitize(parsed_xml)
```
ptrikutam: We're sanitizing/removing tags here because of two reasons:

1. We don't want ElasticSearch to be searching on XML/HTML tags
2. If we're displaying context around our search results, we don't want to display XML/HTML tags, but rather just the content itself.

lib/etl/create_new_resource_record.rb
```diff
@@ -0,0 +1,10 @@
+class CreateNewResourceRecord
+  def write(attributes)
+    record = Resource.create!(attributes)
+
+    ap record.metadata
```
ptrikutam: Can we print out the title here too?

lib/etl/create_new_resource_record.rb
```diff
@@ -0,0 +1,10 @@
...
+    ap record.content.truncate(100)
+
+    record.destroy
```
ptrikutam: Can you remove this?

ptrikutam:
Another note:

On the search results page, creators & publish date aren't showing up: https://cl.ly/1l2r2E2d422o

lib/etl/create_new_resource_record.rb
adarsh: Yes, done in 523e706 and added a little more text so it's clear what's going on.

lib/etl/create_new_resource_record.rb
adarsh: Yep, done in 523e706

adarsh:
Ping @ptrikutam for re-review!

ETL.md
```diff
@@ -0,0 +1,68 @@
...
+│   └── process_all_files_in_local_folder.rb
+├── load
+│   └── create_new_resource_record.rb
+└── transform
+    └── transform_epub.rb
+```
+
+### Rake Tasks
+
+Rake tasks are in `lib/tasks` and namespaced with `etl:`.
```
ptrikutam: Should this be `:etl`?

ETL.md
```diff
@@ -0,0 +1,68 @@
...
+For more information, see the [`kiba`] `README.md` file.
+
+## Usage
+
+Then run the rake task to import the records,
+transform them,
+and load them into the database:
+
+```
+bundle exec rake etl:import_local_epub
```
ptrikutam: Should this perhaps be a more generic rake task? I don't know if this needs to be addressed in this PR but it's specific to an epub, so this command might turn into multiple commands with the addition of more resource types.

Alternatively, should we just add a note requesting developers create a separate rake task for any new resource type they add to the system?

lib/etl/transform/transform_epub.rb
```diff
@@ -0,0 +1,67 @@
+require 'epub/parser'
+
+class TransformEpub
```
ptrikutam: Is it worth adding specs for this? Totally fine if the answer is no.

ETL.md
adarsh: Kinda. I added a clarifying note here: b888f37

lib/etl/transform/transform_epub.rb
adarsh: Eh, the answer is "yes" but maybe "not yet"? I kind of wanted something in place first.

I'll make an issue for this.

adarsh:
> @adarsh actually sorry one more thing -- can you remove the .epub as per @gtourtellot's request?

Yeah, that was done as of d8c6257

Issue #64, opened 2016-12-10
Implement the GroupsUser model linking Group and User


T-Dnzt:
- [x] Create migration for `groups_users` table with appropriate indexes
- [x] Create the `GroupsUser` model with appropriate relationships

PR #65, opened 2016-12-10
Implement the GroupsUser model linking Group and User

T-Dnzt:
[Related Issue](https://github.com/greencommons/commons/issues/64)

This pull request modifies the existing intermediate table between users and groups to allow a new attribute: `admin`. In order for the intermediate model to work correctly with Rails, it was necessary to add an `id`column as well (wasn't needed until now).

The `admin` column holds a boolean defining if a specific user in a specific group is an admin or not. The default is `false`.

The indexes on the table `groups_users` has also been updated. The existing index on `user_id` was left alone, but a compound index was added for `[:group_id, :user_id]`. This compound index will allow fast queries when searching for all the group_users for a specific group or when searching for a specific record using both the group ID and the user ID.


db/migrate/20161210072212_add_id_and_admin_to_groups_users.rb
```diff
@@ -0,0 +1,10 @@
+class AddIdAndAdminToGroupsUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :groups_users, :id, :primary_key
+    add_column :groups_users, :admin, :boolean, null: false, default: false
+
+    # Add a compound index to optimize queries
```
adarsh: Style: Not sure this comment is necessary, but then again, it's in a migration so it's not a huge deal.

T-Dnzt: Yes, you're right... not sure what happened there, I think I'll just remove it.

Issue #60, opened 2016-12-10
Add Group member management functionality


ptrikutam:
### Member Management

Groups can have many admins and many regular members. Admin members are allowed to add/ remove people from the group or grant admin rights to the group, but regular members cannot. For the time being, that's the only difference.

- [x] Add the necessary migrations & changes to the User/Group models to support Admin / Regular members
- [x] Add any necessary routes / controller actions to implement the feature.
- [x] Ensure the route to manage or view the members of a group is: `/groups/id/members`
- [x] Implement an [admin view](https://invis.io/7R9M28BZM) and [regular member](https://invis.io/279M29GJB) view. It should be on the same route, only displaying options on the page if an Admin of the group is logged in. Make sure you use the [style guide](https://greencommons.herokuapp.com/style) as a base for building the UI. The UI doesn't have to be perfect, but should be close to what we see in the wireframes.
- [x] By clicking "Make Admin" below a user's card you should be able to turn a user into an admin for the current group. By Clicking "Remove" you should be able to remove a user from the group. Again, these options are only available to admins.

PR #69, opened 2016-12-10
Member Management Page

T-Dnzt:
[Related Issue](https://github.com/greencommons/commons/issues/60)
Depends on: 
- [PR #68](https://github.com/greencommons/commons/pull/68)
- [PR #67](https://github.com/greencommons/commons/pull/67)
- [PR #66](https://github.com/greencommons/commons/pull/66)
- [PR #65](https://github.com/greencommons/commons/pull/65)

This pull request adds the missing list of members for a group as well as the actions that can be made on them (switch to admin/remove from group). The actions are not using any AJAX/REACT for now.

The `is_admin?`, `find_member`, `add_admin` methods were added to the `Group` model to make it easier to deal with members.

The search feature to add users is also not implemented yet, it will be added in the next PR. Some placeholder texts were used for user name/position/company for now.

members


app/controllers/groups_controller.rb
```diff
@@ -17,6 +17,7 @@ def create
     @group = Group.new(group_params)
 
     if @group.save
+      @group.add_admin(current_user)
```
ptrikutam: Good catch, I would have forgotten about this :)

app/controllers/members_controller.rb
```diff
@@ -0,0 +1,52 @@
+class MembersController < ApplicationController
+  before_action :authenticate_user!
```
ptrikutam: Let's only require authentication for `make_admin`, `remove_admin`, and adding/removing users. Anyone (even people who are logged out) should be able to view a group.

T-Dnzt: Ok, will change that. I will also fix the `GroupsController` to allow anyone to see a group details when I work on the group view page.

PR #70, opened 2016-12-10
Surround the alert/notice with the .row class to avoid strange behaviors

T-Dnzt:
This is a small pull request that adds the `row` class around the alert and error notices. This is needed to avoid some weird floating behavior with the back button.


T-Dnzt:
Well, if everyone agrees that we go with `.slim`, I can create another pull request to switch the remaining `.erb` file. Let me know if that's ok with you @ptrikutam.

ptrikutam:
Sure thing, feel free to start converting over `.erb` to `.slim` files!

T-Dnzt:
Going to merge this. I created [this issue](https://github.com/greencommons/commons/issues/76) for the conversion to `.slim`.

Issue #43, opened 2016-11-30
Deleting Resources has a race condition


adarsh:
> I ran into this issue (after previously manually deleting record 100): https://cl.ly/2x2c1r1I0W26

So that pic is showing errors where the job is failing because it's trying to find record 100 but it doesn't exist. That's a bug - we are asking ES to remove a record from the index which does not exist in the database anymore :(

I'm opening an issue for this. Solution is probably a separate delete job, something like "don't try to rehydrate the AR record when you are deleting it - just use what the job was sent."

T-Dnzt:
@ptrikutam I think we can close this issue.

PR #71, opened 2016-12-10
Fix race condition

adarsh:
Reason for Change
=================
* We were trying to do things with already-deleted records, which is impossible.
* See https://github.com/greencommons/commons/issues/43 for more detail.
* Design-wise, the mistake was trying to make one background job handle several types of indexing work.

Changes
=======
* Alter the single adapter class, `SearchIndex`, to handle adding or removing elements to the search index.
* Add an optional configuration flag, `async: true`, to add/remove from the index asynchronously.
* Change the class interface to all instance methods, since class methods didn't make as much sense.
* Split the single `IndexerJob` class into separate Add and Remove jobs.
* Use `rspec-wait` to ensure indexing background jobs are complete before proceeding.
* Namespace the ElasticSearch index based on the environment to avoid collisions (e.g. indexed records in `:development` not showing up in `:test`).

Minor
=====
* Ignore style checking for the `devise_helper` file, which is vendor code.
* Loosen up the `rspec-rubocop` configs for nesting groups and verified doubles.

https://github.com/greencommons/commons/issues/43


adarsh:
Recommended reviewers: @ptrikutam, @T-Dnzt

lib/etl/load/create_new_resource_record.rb
```diff
@@ -2,9 +2,9 @@ class CreateNewResourceRecord
   def write(attributes)
     record = Resource.create!(attributes)
 
-    ap "Title: #{record.title}"
-    ap 'Metadata: '
-    ap record.metadata
-    ap "Content: #{record.content.truncate(100)}"
+    ap "Title: #{record.title}" # rubocop:disable Rails/Output
```
T-Dnzt: What is Rubocop complaining about here? Any way to change it in the configuration? If we can't, maybe we could add a section to the Readme about the Rubocop offenses that should be disabled with `# rubocop:disable Rails/Output`.

adarsh: The idea is generally to catch your leftover `puts` statements, which is probably generally a good thing (I leave these in accidentally all the time).

Actually, even this `ap` line here is temporary and is used for debugging only, so it's truly a one-off exception.

I'm inclined to leave this as-is for now, but disable this cop immediately the next time it even smells like it's not helpful.

app/models/resource.rb
```diff
@@ -1,5 +1,6 @@
 class Resource < ApplicationRecord
   include Elasticsearch::Model
+  index_name "articles-#{Rails.env}"
```
ptrikutam: Should this be `"resources-#{Rails.env}"`?

app/models/resource.rb
```diff
@@ -20,6 +21,19 @@ class Resource < ApplicationRecord
...
+  after_save do
+    SearchIndex.new(
+      model_name: self.class.name,
+      id: id,
+      async: true,
+    ).add
+  end
+
+  after_destroy do
+    SearchIndex.new(
```
ptrikutam: Would it make sense to have `RemoveFromIndexJob` directly here instead of having the path go `SearchIndex` -> `RemoveFromIndexJob` -> `SearchIndex`?

Not saying my suggestion is better -- but it feels a bit confusing to have one class refer to another, no? Alternatively, would it make sense to have the code to actually add or remove the item from the index to exist within `AddToIndexJob` and `RemoveFromIndexJob`? Curious what you think is best for this part.

adarsh: > Would it make sense to have RemoveFromIndexJob directly here instead of having the path go `SearchIndex` -> `RemoveFromIndexJob` -> `SearchIndex`?

Yeah I debated this while writing it and settled on the `async: true` flag to hide the job stuff away. However, since you raised it, I'm going to call this "confusing" and just create the new job directly from `Resource` a new PR.

> Alternatively, would it make sense to have the code to actually add or remove the item from the index to exist within AddToIndexJob and RemoveFromIndexJob?

This is what I was hoping to avoid with `SearchIndex` -  abstract as much of the ElasticSearch-specific stuff in one utility class as I could. That's actually why I set up the runaround described above.

app/models/resource.rb
adarsh: Whoops! Yes.

app/models/resource.rb
adarsh: Opened a fix PR here: https://github.com/greencommons/commons/pull/78

app/models/resource.rb
adarsh: Fix proposed in https://github.com/greencommons/commons/pull/78

PR #78, opened 2016-12-12
Reduce logical complexity for search index jobs

adarsh:
Reason for Change
=================
* See comment at https://github.com/greencommons/commons/pull/71#discussion_r91873393 :

> Would it make sense to have `RemoveFromIndexJob` directly here instead of having the path go `SearchIndex` -> `RemoveFromIndexJob` -> `SearchIndex`?

Yes.

Changes
=======
* Call the job enqueuing directly from the ActiveRecord model.
* Make the callback flag check and logging class-level methods.
* BUG: Correct index-name configuration error.

Minor
-----
* Set an RSpec file for storing metadata on the last suite run. This is useful for running flags like `--last-failure`.

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -1,6 +1,6 @@
 class Resource < ApplicationRecord
   include Elasticsearch::Model
-  index_name "articles-#{Rails.env}"
+  index_name SearchIndex.index_name(self)
 
   RESOURCE_TYPES = {
     article: 0,
@@ -21,19 +21,11 @@ class Resource < ApplicationRecord
 
   validates :title, :resource_type, presence: true
 
-  after_save do
-    SearchIndex.new(
-      model_name: self.class.name,
-      id: id,
-      async: true,
-    ).add
+  after_commit on: [:create, :update] do
+    AddToIndexJob.perform_async(self.class.name, id)
   end
 
-  after_destroy do
-    SearchIndex.new(
-      model_name: self.class.name,
-      id: id,
-      async: true,
-    ).remove
+  after_commit on: [:destroy] do
+    RemoveFromIndexJob.perform_async(self.class.name, id)
   end
 end
--- a/app/service/search_index.rb
+++ b/app/service/search_index.rb
@@ -1,60 +1,51 @@
 class SearchIndex
-  def initialize(model_name:, id:, async: false)
+  def self.index_name(record)
+    "#{record.class.name.pluralize.downcase}-#{Rails.env}"
+  end
+
+  def self.log_elasticsearch_warning(message)
+    Rails.logger.tagged('ELASTICSEARCH') do
+      Rails.logger.warn(message)
+    end
+  end
+
+  def initialize(model_name:, id:)
     @model_name = model_name
     @id = id
-    @async = async
   end
 
   def add
-    if async
-      add_to_index_asynchronously
-    else
+    if search_index_callbacks_enabled?
       record = model_name.constantize.find(id)
       record.__elasticsearch__.index_document
+    else
+      log_callback_warning
     end
   end
 
   def remove
-    if async
-      remove_from_index_asynchronously
-    else
+    if search_index_callbacks_enabled?
       Elasticsearch::Model.client.delete(
         index: model_name.constantize.index_name,
         type: model_name.downcase,
         id: id,
       )
+    else
+      log_callback_warning
     end
   end
 
   private
 
-  attr_reader :async, :id, :model_name
-
-  def add_to_index_asynchronously
-    if search_index_callbacks_enabled?
-      AddToIndexJob.perform_async(model_name, id)
-    else
-      log_warning
-    end
-  end
-
-  def remove_from_index_asynchronously
-    if search_index_callbacks_enabled?
-      RemoveFromIndexJob.perform_async(model_name, id)
-    else
-      log_warning
-    end
-  end
+  attr_reader :id, :model_name
 
   def search_index_callbacks_enabled?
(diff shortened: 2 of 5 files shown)


app/models/resource.rb
```diff
@@ -1,6 +1,6 @@
 class Resource < ApplicationRecord
   include Elasticsearch::Model
-  index_name "articles-#{Rails.env}"
+  index_name "#{self.class.name.downcase}-#{Rails.env}"
```
T-Dnzt: Good stuff. We could pluralize it to follow the previous style. I'm also in the habit of making logic "paths" for caches/indexes as in using `/development/resources`, but that's just a  personal preference - totally ok with using model-env and I will use the same format if I need to in the rest of the project.

app/service/search_index.rb
```diff
@@ -1,60 +1,35 @@
...
-  def remove
-    if async
-      remove_from_index_asynchronously
-    else
-      Elasticsearch::Model.client.delete(
-        index: model_name.constantize.index_name,
-        type: model_name.downcase,
-        id: id,
+  def self.log_callback_warning
+    Rails.logger.tagged('ELASTICSEARCH') do
```
T-Dnzt: I notice these three same lines are repeated 3 times with only the warning content changing. Maybe it would be nice to have a single method that deals with logging to which we could pass the line to log? Maybe a class method on `SearchIndex` or a very small and simple ElasticSearch logger class where we could easily add more logging options later on (if we need it)?

A quick example for reference:

```ruby
def self.log_es_warning(warning)
  Rails.logger.tagged('ELASTICSEARCH') do
    Rails.logger.warn(warning)
  end
end
```

app/models/resource.rb
```diff
@@ -22,18 +22,18 @@ class Resource < ApplicationRecord
   validates :title, :resource_type, presence: true
 
   after_save do
-    SearchIndex.new(
-      model_name: self.class.name,
-      id: id,
-      async: true,
-    ).add
+    if SearchIndex.search_index_callbacks_enabled?
```
ptrikutam: This is closer to what we want, but still feels like there's some tight coupling between the `SearchIndex` class and the `AddToIndexJob`/`RemoveFromIndexJob` classes.

In my mind, I'd like to see a single direction of the flow

`Resource#after_save -> AddToIndexJob#perform_async -> SearchIndex#add`

So either `AddToIndexJob` or `SearchIndex` should handle checking `search_index_callbacks_enabled?` and deciding to log something or not. I'm leaning towards having `SearchIndex` make that decision since otherwise we'd have duplicate code in both `AddToIndexJob` and `RemoveFromIndexJob`, and they'd both be calling `SearchIndex#log_callback_warning` anyway.

app/models/resource.rb
adarsh: I'm totally into using slashes in the index names as long at they are allowed characters. It seems like [maybe they are not?](https://github.com/elastic/elasticsearch/issues/6736#issue-37167454).

Either way, I've pluralized in this commit: e0c9663.

adarsh: Actually, I've extracted this to `SearchIndex` in 8f8b584 also - this will be useful elsewhere.

app/models/resource.rb
adarsh: Yeah I totally agree. Reworked it here: 75c0c70

adarsh:
@ptrikutam ready for re-review.

app/models/resource.rb
```diff
@@ -1,6 +1,6 @@
 class Resource < ApplicationRecord
   include Elasticsearch::Model
-  index_name "articles-#{Rails.env}"
+  index_name SearchIndex.index_name(self)
```
ptrikutam: 👍

spec/service/search_index_spec.rb
```diff
@@ -20,42 +40,26 @@
       end
     end
 
-    context 'when asked to be performed asynchronously' do
-      it 'enqueues a job to include the record in the search index' do
-        allow(AddToIndexJob).to receive(:perform_async)
-        record = build_stubbed(:resource)
+    context 'if callbacks are disabled', search_indexing_callbacks: false do
```
ptrikutam: Had no idea you could do this. Legit, this test is well written

ptrikutam:
Further steps:

1. Set up Resources in DB & verify they have associated index documents. `Resource.count == 100`
1. Reset Sidekiq database.
1. Call `Resource.destroy_all`. In Sidekiq, 100 jobs have bee successfully processed. This is expected.
1. Call `bundle exec rake db:seed`. After all jobs have finished , Sidekiq has 423 jobs processed, 63 failed.

Basically, calling `rake db:seed` seems to trigger 386 jobs. Of those, 63 fail.

ptrikutam:
@adarsh went over the details of how Sidekiq is defined, and this is no longer an issue. Apparently the problem is caused because Sidekiq is actually executing faster than the database: https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#cannot-find-modelname-with-id12345

Another point of confusion: "Failed" in the Sidekiq web UI refers to the number of failed _attempts_, not the number of failed _jobs_, which would fall under the "Dead" count. Therefore, the numbers make more sense as there are a number of tries that failed, then equivalent number of tries that succeed. 

This PR is good to go, but if @adarsh is so inclined, he can implement the `:after_commit` suggestion in the Sidekiq FAQ link above :)

app/models/resource.rb
T-Dnzt: Oh ok, didn't know these characters were not allowed. At least, now I know we're both into slashes! And good call on extracting it, going to be super easy to reuse like this.

Issue #79, opened 2016-12-13
Add Group Summary card to /style


ptrikutam:
Can you add an example of the group summary card (and a code snippet of its usage in Rails) on the `/style` page? I'd like to keep an up-to-date list of all the reusable components we create.

T-Dnzt:
Would you like to have syntax coloration for the code examples?

T-Dnzt:
Is [this](https://github.com/greencommons/commons/pull/82) what you had in mind?

ptrikutam:
Going to respond to your Qs on the PR itself.

PR #82, opened 2016-12-14
Update style to include the summary cards for groups and groups_users

T-Dnzt:
[Related Issue](https://github.com/greencommons/commons/issues/79)

This pull request adds some examples of the summary cards usage to the `/style` page. I included the [highlight.js](https://highlightjs.org) library directly in the style view because we don't need it anywhere else. Also, there is no support for slim apparently so I'm not sure if we want to keep it or not... 

To have some data to show, I used the factories from the project. I don't think we want to mess with the DB but we could maybe use the first group if any and fallback on the factory (which has some disadvantages and might fail in the future since we don't have a persisted object with an `id` for example).

Here are screenshots of the addition:

screen shot 2559-12-14 at 5 04 05 pm
screen shot 2559-12-14 at 5 04 09 pm


app/views/pages/style.html.erb
```diff
@@ -1,9 +1,90 @@
+<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.9.0/styles/default.min.css">
+<script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.9.0/highlight.min.js"></script>
+<script>hljs.initHighlightingOnLoad();</script>
+
 <div class="text-center">
```
ptrikutam: Small nit -- can you place this below the additions you've made? I want to keep the original theme styles and our custom components separate.

ptrikutam:
Hey @T-Dnzt -- Syntax highlighting isn't required. Just a `<pre>` block should be enough, I think. Highlighting support was the motivation to keep both `.html.erb` and `.slim`, we can get rid of the `.html.erb` files.

app/views/pages/style.html.erb
T-Dnzt: Oh certainly, I will change it.

T-Dnzt:
@ptrikutam Just removed the `.erb` version, we don't need it anymore. I also removed the highlighting and created a separate section title "Green Commons Visual Components" so people can easily understand what's part of the theme and what's part of Green Commons.

Issue #75, opened 2016-12-12
Ensure only group admins can update or delete a group


T-Dnzt:
There are currently no permission checks in the `GroupsController` for `update` and `destroy`. These actions should only be performed by group admins, and therefore, we need to ensure that the `current_user` is an admin before allowing the changes.

Issue #61, opened 2016-12-10
Add group view page


ptrikutam:
Based on [the wireframes for the Group View page](https://invis.io/BH8GKH38D#/185307852_GROUP_-_View), please start coding up a skeleton for the Group view.

Some points about functionality:

- A `Group` can have many `List`s. Each of these `List`s will have many resources -- anyone in a group can modify a `List` that belongs to the `Group`.
- In the "What's New" section, we want to display recent items that have been added to any `List` owned by the `Group`. This might require creating some kind of helper method / scope on `Group` to access all of its `Resource`s through its various `List`s. Open to discussion on the best way to approach this.
- Clicking the "See More" link should display the rest of the Group's metadata, such as the description, website, created date, etc.
- Anyone can view a group, even if they're not a member
- Anyone can view a group's member list, even if they're not a member
- The ["Leave Group" button](https://cl.ly/403W1A0V3M41) should only appear if you're a member of the group
- Please leave out the "Add to List" button for now.

T-Dnzt:
This is the summary card for a group:

![screen shot 2559-12-14 at 8 18 18 pm](https://cloud.githubusercontent.com/assets/2647689/21183444/849beaa6-c23a-11e6-82a2-d27dbb65c366.png)

And this for a resource:

![screen shot 2559-12-14 at 8 18 21 pm](https://cloud.githubusercontent.com/assets/2647689/21183446/8ad1f2c6-c23a-11e6-88fe-521f675dc5f1.png)

As you can see the tags/metadata are not presented in the same order on both. Do we want to unify this or should I implement the two different styles?

T-Dnzt:
I have a skeleton on the view and I'm wondering about the button colors. Either I can have them all blue, or use different colors depending on the action type as shown in the screenshot below. Blue for GET, green for POST and red for DELETE:

![screen shot 2559-12-14 at 9 08 54 pm](https://cloud.githubusercontent.com/assets/2647689/21185011/8ac9264e-c241-11e6-94ae-0cdd4955e588.png)

T-Dnzt:
A basic version of the group view page is available in this [PR](https://github.com/greencommons/commons/pull/83). There's nothing dynamic yet and a lot of fake data.

ptrikutam:
> As you can see the tags/metadata are not presented in the same order on both. Do we want to unify this or should I implement the two different styles?

Good catch. Feel free to use your best judgement and unify the styles as you see fit.

[...]

PR #83, opened 2016-12-14
Implement basic Group view page

T-Dnzt:
# Introduction

This pull request introduces the group view page showing all the information about a specific group as well as the related resources, lists and comments. 

It also includes the addition of Pundit to handle permissions for the various actions of the `GroupsController` and `MembersController`. Note that the permissions are currently shared and only defined in the `GroupPolicy` class. Inside the `MembersController`, all actions adding/changing/removing members are associated with updating a group (only admins can do it). The alternative would be to change the Pundit helper method to take three arguments instead of two (user, group_user, group). 

# Related issues

- [Issue #61](https://github.com/greencommons/commons/issues/61) 
- [Issue #75](https://github.com/greencommons/commons/issues/75)

# Screenshots

screen shot 2559-12-16 at 12 11 03 pm

screen shot 2559-12-23 at 8 52 21 am


Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'sass-rails', '~> 5.0'
 gem 'sidekiq'
 gem 'slim'
 gem 'uglifier', '>= 1.3.0'
+gem 'pundit'
```
adarsh: I approve of this gem 😉

app/controllers/members_controller.rb
```diff
@@ -27,18 +28,33 @@ def create
...
 
+  def leave
+    @group_current_user = @group.find_member(current_user)
+    authorize @group_current_user
+
+    if !@group_current_user.admin? || @group.admin_count > 1
+      @group_current_user.destroy
+      redirect_to @group, notice: 'You are no longer a member of this group.'
+    else
+      redirect_to @group, notice: 'You cannot leave this group because you are the only admin.'
```
ptrikutam: Great catch :)

app/models/group.rb
```diff
@@ -5,11 +5,12 @@ class Group < ApplicationRecord
...
   has_many :users, through: :groups_users
   has_many :lists, as: :owner
+  has_many :resources, through: :lists
 
   validates :name, presence: true
 
-  def resources
-    lists.map(&:resources).flatten.uniq
+  def latest_resources(limit: 4)
+    resources.sort_by_created_at.distinct.limit(limit)
```
ptrikutam: I dig it. If I'm not mistaken, this change will move these operations to the database level, and not do them in memory, correct?

spec/policies/group_policy.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    it { is_expected.to forbid_action(:index)   }
+    it { is_expected.to forbid_action(:new)     }
+    it { is_expected.to forbid_action(:create)  }
+    it { is_expected.to forbid_action(:update)  }
+    it { is_expected.to forbid_action(:edit)    }
+    it { is_expected.to forbid_action(:destroy) }
+  end
+
+  context 'when regular group member' do
```
ptrikutam: Should there be another spec here for when a person is logged in, but not a member?

I would imagine the permitted actions are then: `:show`, `:index`, `:new`, `:create`.

Unless I'm confusing what those policies are meant for... What do you think?

app/views/groups/show.html.slim
```diff
@@ -1,10 +1,142 @@
...
+        = render 'shared/components/btn_count', path: '#', 
+                                                text: 'ADD TO LIST', 
+                                                label: @group.lists.count,
+                                                classes: 'resource-actions__button--right',
+                                                style: 'primary'
+        
+        - if @group_current_user                                 
+          = link_to 'LEAVE GROUP', leave_group_members_path(@group), 
+                                   method: :delete, 
+                                   data: { confirm: 'Are you sure?' },
```
ptrikutam: Can we update this to "Are you sure you want to leave this group?"

app/models/group.rb
T-Dnzt: Correct, it will all happen in one SQL query.

spec/policies/group_policy.rb
T-Dnzt: I think you're right, indeed ;) Let me look into it.

T-Dnzt: And thanks for pointing that out, made me realize I'm missing the `_spec` suffix in the filename...

app/controllers/members_controller.rb
T-Dnzt: Without this, we can say hello to orphan groups :D

Issue #80, opened 2016-12-14
Update seeds.db to use FactoryGirl


ptrikutam:
Reduces the number of places we need to update when we update a model.

Please try to follow a similar structure:

- Create multiple users with 1 - 2 lists (use a random sampling of 10 - 50 resources for each).
- Create multiple groups with multiple users.
- For each group, create 3 - 5 lists (with a random sampling of 30 - 80 resources for each list)

PR #88, opened 2017-01-04
Update seeds with FactoryGirl

T-Dnzt:
[Related issue: #80](https://github.com/greencommons/commons/issues/80)

This pull request comes with a new version of the `seed.rb` file, using `FactoryGirl` to create the records. 

It follows the structure presented below:

- Create multiple users with 1 - 2 lists (use a random sampling of 10 - 50 resources for each).
- Create multiple groups with multiple users.
- For each group, create 3 - 5 lists (with a random sampling of 30 - 80 resources for each list)

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -3,11 +3,12 @@ users_count = 20
 resources_count = 400
 groups_count = 10
 
+p 'Creating admin user with email \'admin@greencommons.org\' and password \'thecommons\'...'
+admin = FactoryGirl.create(:user, email: 'admin@greencommons.org',
+                                  password: 'thecommons')
+
 p "Creating #{users_count} users..."
-users_count.times do |n|
-  FactoryGirl.create(:user, email: "user#{n}@greencommons.org",
-                            password: 'thecommons')
-end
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,76 +1,38 @@
...
-    password: 'thecommons'
-  )
+# config
+users_count = 20
+resources_count = 400
+groups_count = 10
+
+p "Creating #{users_count} users..."
+users_count.times do |n|
+  FactoryGirl.create(:user, email: "user#{n}@greencommons.org",
```
ptrikutam: What about using FactoryGirl's `create_list` for this one. No need to specify email or password, I'm fine with those being randomized values from Faker generated by the factory itself. What do you think?

T-Dnzt: My only concern with this is that it doesn't give a user for the developer to play with. I've personally been using user0@greencommons.org and thecommons to login since I started working on the app. Should we create a separate one and use create_list for the other users?

ptrikutam: Good idea. Let's create an `admin@greencommmons.org` and use `create_list` for the rest.

Issue #74, opened 2016-12-12
Add Capybara tests for the Group Functionality


T-Dnzt:
There are currently no complete feature tests to ensure that the various parts of the Group Functionality work as expected. 

Here are the happy path scenarios that have to be implemented as acceptance tests:

- [x] `scenario 'users can create a group'`
- [x] `scenario 'users can update a group'`
- [x] `scenario 'users can add members'`
- [x] `scenario 'users can remove members'`
- [x] `scenario 'users can make other members admins'`
- [x] `scenario 'users can remove admin from other members'`

Some of these scenarios might get merge together to have faster tests.

ptrikutam:
Great idea. Please do flesh this issue out once you're done w/ the Group PR.

Issue #53, opened 2016-12-02
Create capybara tests to verify auth flows


ptrikutam:
There's no need to test the registration / auth controllers, but in case we've messed something up on the views end we should add a few integration tests to verify the auth flow works as we expect.

PR #89, opened 2017-01-05
Add feature test for Authentication

T-Dnzt:
[Related Issue: #74](https://github.com/greencommons/commons/issues/74) and [#53](https://github.com/greencommons/commons/issues/53).

This is a preliminary PR for issue #74. Since users must be logged in to manage groups, I thought it would be nice to have some happy path acceptance tests for the login/logout. I know Devise is handling it, but it's always nice to ensure that we won't have a regression one day by messing with the login form for example.

Implementing this also led me to wonder if we want to stub the logged in user in the acceptance tests for the groups management (faster), or if we should run the user through the login page every time (slower). I like the latter more, since it acts more like a real user, but I'm worried about the speed cost.

PS: These tests pointed out that `autocomplete: off` was duplicated in the `session/new` view.

### DIFF
--- a/app/views/devise/sessions/new.html.slim
+++ b/app/views/devise/sessions/new.html.slim
@@ -11,7 +11,7 @@
       .form-group
         = f.label :password, class: 'col-lg-2 control-label'
         .col-lg-10
-          = f.password_field :password, autocomplete: 'off', class: 'form-control', placeholder: 'Password', autocomplete: 'off'
+          = f.password_field :password, autocomplete: 'off', class: 'form-control', placeholder: 'Password'
       .form-group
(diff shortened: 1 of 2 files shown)


ptrikutam:
> Implementing this also led me to wonder if we want to stub the logged in user in the acceptance tests for the groups management (faster), or if we should run the user through the login page every time (slower). I like the latter more, since it acts more like a real user, but I'm worried about the speed cost.

I agree with you. For the time being, we're not really slowed down by our specs, and a bit of a slower spec is fine if it's doing a thorough job of testing critical paths. I think if the feature specs really start to take forever, we can switch to stubbing the user vs. actually going through the log in flow.

PR #93, opened 2017-01-08
Groups feature spec

T-Dnzt:
[Related Issue: #74](https://github.com/greencommons/commons/issues/74)

This pull request adds acceptance/feature tests for the group management system. The following are being tested with the four tests included in this PR:

## Group Management

- [x] `scenario 'users can create a group'`
- [x] `scenario 'users can update a group'`
- [x] `scenario 'users can add members'`
- [x] `scenario 'users can remove members'`
- [x] `scenario 'users can make other members admins'`
- [x] `scenario 'users can remove admin from other members'`

## Group Interactions

- [x] Non-members can join an existing group
- [x] Ensure that regular members (not admin) can't edit group
- [x] Ensure that members can leave a group
- [x] Ensure that non-admin members can't edit members list

Two HTML tags were added in the form to add new members to a group in order to make testing it easier.


spec/feature/user_manage_groups_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'rails_helper'
+
+RSpec.feature 'Managing groups' do
+  scenario 'users can create a group' do
+    feature_login
+    group = build(:group)
+
+    click_link 'Create Group'
```
ptrikutam: Strange that you don't have to click on the plus sign first and this works. Do you know why that is?

T-Dnzt: The `Create Group` link is inside a hidden `div`, but it's still present on the HTML page which is why Capybara can locate it anyway.

T-Dnzt:
I can add these tests, sure! I wanted to avoid adding too many acceptance tests to keep the suite fast, but as you said in the other PR, it's still fast enough for now. We can always speed things up later if needed.

Issue #98, opened 2017-01-09
Search Results Page


ptrikutam:
We want to build out the Search results page: https://invis.io/BH8GKH38D#/185307864_SEARCH

This will involve a few different things:

- [x] Taking the search term entered and searching across `Group`, `Resource`, and `List` to begin with. The search page should return a list of all items combined together.
- [x] If they don't already exist, we need to create summary cards for different search result items.
- [x] Building out the left-hand filter list. We should discuss what types of filters actually belong in this section. At the minimum, we should be able to filter by Group, Resource, and List.
- [x] Allow sorting by Date or "Relevance" (see "Stackscore / Relevancy" below)
- [x] Ensure we can paginate search results

## Stackscore / Relevancy

@dweinberger wants to assign a "Stackscore" or "Relevancy" number to every search result we get. This can be a number between 0 and 100 or something similar, but ultimately is a "smart" determination of what an item's relevancy is to what you've searched for. Initially, we can make this very simple, but we should build a stub or foundation to expand and make this more sophisticated as we go along. I'm happy to walk through this via a audio / video chat.

## "You may also like"

Not pictured in the wireframes, we want to include a small section below the search results that provides suggestions on other things a user may be interested in. Initially, we can keep it simple, and the recommendations can be rather uninteresting, but it'd be good to set up a way to see "related items" based on a search term or an actual object in the database. This is somewhat similar in idea to "Stackscore / Relevancy," so it's worth discussing what makes sense here.

**NOTE**: This is a rather large task, so if it makes sense, feel free to break this up into smaller issues and re-order as you see fit.

T-Dnzt:
I guess we can close this issue :) If I forgot anything, let me know!

PR #106, opened 2017-01-13
Setup ES indices for Groups and Lists

T-Dnzt:
Related issue: #98 

> Taking the search term entered and searching across Group, Resource, and List to begin with. The search page should return a list of all items combined together.

This pull request turns `Group` and `List` into models indexed with ElasticSearch. The logic was extracted from the `Resource` model and moved into a concern (`Indexable`) for easy reuse. 

New rake tasks were also added to reset/create the indices for the new models (with a fallback if the `Model.import` method fails). 

Finally, the tests checking if a model is correctly indexed on `create` / `delete` were moved into a shared example and included in each of the models mentioned above.

### DIFF
--- a/lib/tasks/elasticsearch.rake
+++ b/lib/tasks/elasticsearch.rake
@@ -1,15 +1,42 @@
 namespace :elasticsearch do
-  desc 'Deletes the "resource" index and regenerates it with all records currently in Resource'
-  task reset_resource_index: :environment do
+
+  def reset_index(klass)
     client = Elasticsearch::Client.new(
       url: ENV.fetch('BONSAI_URL', 'http://localhost:9200'), log: true
     )
 
-    if client.indices.exists? index: Resource.index_name
-      client.indices.delete index: Resource.index_name
+    if client.indices.exists? index: klass.index_name
+      client.indices.delete index: klass.index_name
     end
 
-    Resource.__elasticsearch__.create_index!
-    Resource.import
+    klass.__elasticsearch__.create_index!
+
+    begin
+      klass.import
+    rescue Faraday::ConnectionFailed => e
+      ap e
+      ap 'Indexing records one at a time...'
+      klass.all.each { |r| r.__elasticsearch__.index_document }
+    end
+  end
+
+  desc 'Deletes the "resource", "group" and "list" indices and regenerates it with all records'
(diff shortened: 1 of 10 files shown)


lib/tasks/elasticsearch.rake
```diff
@@ -1,15 +1,41 @@
...
+      client.indices.delete index: klass.index_name
     end
 
-    Resource.__elasticsearch__.create_index!
-    Resource.import
+    klass.__elasticsearch__.create_index!
+
+    begin
+      klass.import
+    rescue Faraday::ConnectionFailed => e
```
ptrikutam: Interesting. Is Faraday used to connect to ES?

Does indexing one at a time resolve this? More curious than anything else.

lib/tasks/elasticsearch.rake
```diff
@@ -1,15 +1,41 @@
...
+      ap e
+      ap 'Indexing records one at a time...'
+      klass.all.each { |r| r.__elasticsearch__.index_document }
+    end
+  end
+
+  desc 'Deletes the "resource", "group" and "list" indices and regenerates it with all records'
+  task reset_all_indices: :environment do
+    [Resource, Group, List].each { |klass| reset_index(klass) }
```
ptrikutam: 👏

spec/models/list_spec.rb
```diff
@@ -17,4 +17,31 @@
     it { is_expected.to belong_to(:owner) }
     it { is_expected.to have_and_belong_to_many(:resources) }
   end
+
+  describe 'Callbacks' do
```
ptrikutam: Do we still need this if we have the shared examples for "indexable"?

lib/tasks/elasticsearch.rake
T-Dnzt: Yes, Faraday is used to make the HTTP calls to interact with Elastic Search. Indexing one at a time seems totally fine. This was a quick fix until we've had more time to investigate the issue (and if it's even worth our time).

spec/models/list_spec.rb
T-Dnzt: No indeed, I thought I removed it but apparently I forgot. Will fix that.

PR #109, opened 2017-01-16
Add support for ETL of JSON resource and group data.

gtourtellot:
* ETL.md shows rakes to use and new ETL files.
* Groups and Resources JSON schema are compatible with and extend db/schema.rb defs.
* Local data in lib/etl/data are useful for simple tests.
* Ingestion of JSON from S3 is supported.
* S3 JSON data derives from sources including EPUBs augmented with ONIX data.

### DIFF
--- a/ETL.md
+++ b/ETL.md
@@ -35,12 +35,20 @@ For more information, see the [`kiba`] `README.md` file.
 
 ## Usage
 
-Then run the rake task to import the records,
+Then run the rake tasks to import the records,
 transform them,
 and load them into the database:
 
 ```
-bundle exec rake etl:import_local_epub
+with local test files:
+
+bundle exec rake etl:import_local_json_resources
+bundle exec rake etl:import_local_json_groups
+
+or with files from S3:
+
+bundle exec rake etl:import_s3_json_resources
+bundle exec rake etl:import_s3_json_groups
 ```
 
 ## Development
@@ -51,16 +59,24 @@ kiba ETL source files are in `lib/etl/`
 
 ```
 ├── common.rb
-├── convert-islandpress.etl
+├── process_local_json_resources.etl
+├── process_local_json_groups.etl
+├── process_s3_json_resources.etl
+├── process_s3_json_groups.etl
 ├── data
-│   ├── sample_island_press_1.epub
-│   └── sample_island_press_2.epub
+│   ├── good_resource_array.json
+│   └── good_group_array.json
 ├── extract
-│   └── process_all_files_in_local_folder.rb
+│   └── process_files_in_s3_bucket_folders.rb
 ├── load
-│   └── create_new_resource_record.rb
-└── transform
-    └── transform_epub.rb
+│   ├── create_new_resource_records.rb
+│   └── create_new_group_records.rb
+├── transform
+│   └── transform_json.rb
+└── schema
+    ├── resource_array_schema.json
+    └── group_array_schema.json
+    
 ```
 
 ### Rake Tasks
--- a/lib/etl/data/bad_group_array.json
+++ b/lib/etl/data/bad_group_array.json
@@ -1,0 +1,17 @@
+[
+	{
+		"namEE": "Ima Invalid Group",
+		"long_description": "Org for goodness sake",
+		"url": "http://www.goodnesssake.org",
+		"metadata": {
+		}
+	},
+	{
+		"name": "Ima Good One",
+		"long_description": "All day, every day, combating polluting corporations.",
+		"email": "support@greenfighingtback.org",
+		"metadata": {
+			"founded": "2017-01-01"
+		}
+	}
+]
--- a/lib/etl/extract/process_files_in_s3_bucket_folders.rb
+++ b/lib/etl/extract/process_files_in_s3_bucket_folders.rb
@@ -1,0 +1,52 @@
+require 'aws-sdk'
+require 'tempfile'
+
+class ProcessFilesInS3BucketFolders
+  def initialize(bucket_name, prefix_folders, file_ext)
+    @bucket_name = bucket_name
+    @prefix_folders = prefix_folders
+    @file_ext = file_ext
+  end
+
+  def each
+    filtered_files.each do |file_key|
+      Tempfile.open('s3data', dir, binmode: true) do |tempfile|
+        s3.get_object(
+          response_target: tempfile.path,
+          bucket: bucket_name,
+          key: file_key,
+        )
+
+        yield(tempfile, file_key)
+      end
+    end
+  end
+
+  private
+
+  attr_reader :bucket_name, :prefix_folders, :file_ext
+
+  def dir
+    "#{Rails.root}/tmp"
+  end
+
+  def filtered_files
+    file_keys.grep(/#{file_ext}/)
+  end
+
+  def file_keys
+    arr = []
+    @prefix_folders.each do |prefix_folder|
+      arr += s3.list_objects(bucket_location(prefix_folder)).contents.map(&:key)
+    end
+    arr
+  end
+
+  def s3
+    @_s3 = Aws::S3::Client.new
+  end
+
+  def bucket_location(prefix_folder)
+    { bucket: bucket_name, prefix: prefix_folder }
+  end
+end
--- a/lib/etl/load/create_new_group_records.rb
+++ b/lib/etl/load/create_new_group_records.rb
@@ -1,0 +1,32 @@
+class CreateNewGroupRecords
+  def write(attributes_list)
+    @attributes_list = attributes_list
+
+    attributes_list.each do |attributes|
+      puts "#{attributes["name"]}"
+      
+      if existing_record attributes
+        ap "Name already in database: #{name attributes}" # rubocop:disable Rails/Output
+      else
+        record = Group.create!(attributes)
+
+        ap "Name: #{record.name}" # rubocop:disable Rails/Output
+        ap 'Metadata: ' # rubocop:disable Rails/Output
+        ap record.metadata # rubocop:disable Rails/Output
+        ap record.long_description # rubocop:disable Rails/Output
+      end
+    end
+  end
+
+  private
+
+  attr_reader :attributes_list
+
+  def existing_record(attributes)
+    Group.find_by(name: name(attributes))
+  end
+
+  def name(attributes)
+    attributes.fetch(:name)
+  end
+end
--- a/lib/etl/load/create_new_resource_records.rb
+++ b/lib/etl/load/create_new_resource_records.rb
@@ -1,0 +1,32 @@
+class CreateNewResourceRecords
+  def write(attributes_list)
+    @attributes_list = attributes_list
+
+    attributes_list.each do |attributes|
+      puts "#{attributes["title"]}"
+      
+      if existing_record attributes
+        ap "Title already in database: #{title attributes}" # rubocop:disable Rails/Output
+      else
+        record = Resource.create!(attributes)
+
+        ap "Title: #{record.title}" # rubocop:disable Rails/Output
+        ap 'Metadata: ' # rubocop:disable Rails/Output
+        ap record.metadata # rubocop:disable Rails/Output
+        ap "Content: #{record.content.truncate(100)}" # rubocop:disable Rails/Output
+      end
+    end
+  end
+
+  private
+
+  attr_reader :attributes_list
+
+  def existing_record(attributes)
+    Resource.find_by(title: title(attributes))
+  end
+
+  def title(attributes)
+    attributes.fetch(:title)
+  end
+end
--- a/lib/etl/transform/transform_json.rb
+++ b/lib/etl/transform/transform_json.rb
@@ -1,0 +1,38 @@
+require 'json'
+require 'json-schema'
+
+class TransformJson
+  def process(input_json)
+    @input_json = input_json
+    pl = parsed_list
+    JSON::Validator.validate(schema_fname, pl)
+    pl
+    
+  rescue => error
+    ap "Error opening json file: #{error}"
+  end
+
+  private
+
+  attr_reader :input_json, :schema_fname
+  
+  def schema_fname
+    raise NotImplementedError.new("#{self.class.name}#area is an abstract method.")
+  end
+  
+  def parsed_list
+    JSON.parse(File.read(input_json), :symbolize_names => true)
+  end
+end
+
+class TransformJsonResources < TransformJson
+  def schema_fname
+    "lib/etl/schema/resource_array_schema.json"
+  end
+end
+
+class TransformJsonGroups < TransformJson
+  def schema_fname
+    "lib/etl/schema/group_array_schema.json"
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem 'kiba', '~> 1.0.0'
 gem 'epub-parser'
 gem 'awesome_print'
 gem 'factory_girl_rails'
+gem 'json-schema'
 
 group :development do
   gem 'bummr'
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ gem 'pundit'
 gem 'kiba', '~> 1.0.0'
 gem 'epub-parser'
 gem 'awesome_print'
+gem 'factory_girl_rails'
 
 group :development do
   gem 'bummr'
@@ -39,7 +40,6 @@ group :development, :test do
   gem 'climate_control'
   gem 'database_cleaner'
   gem 'dotenv-rails'
-  gem 'factory_girl_rails'
   gem 'pry-byebug'
   gem 'rspec-rails', '~> 3.5'
   gem 'rspec-wait'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -165,6 +165,8 @@ GEM
       railties (>= 4.2.0)
       thor (>= 0.14, < 2.0)
     json (2.0.2)
+    json-schema (2.7.0)
+      addressable (>= 2.4)
     kiba (1.0.0)
     listen (3.0.8)
       rb-fsevent (~> 0.9, >= 0.9.4)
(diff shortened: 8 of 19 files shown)


lib/etl/extract/process_files_in_s3_bucket_folders.rb
```diff
@@ -0,0 +1,52 @@
+require 'aws-sdk'
+require 'tempfile'
+
+class ProcessFilesInS3BucketFolders
+  def initialize(bucket_name, prefix_folders, file_ext)
+    @bucket_name = bucket_name
+    @prefix_folders = prefix_folders
```
ptrikutam: Out of curiosity, why is this an array? It seems like you're only passing arrays with one value in the lower areas, so wouldn't it simplify this class to keep it as an individual folder vs. multiple folders?

lib/etl/transform/transform_json.rb
```diff
@@ -0,0 +1,38 @@
+require 'json'
+require 'json-schema'
+
+class TransformJson
+  def process(input_json)
+    @input_json = input_json
+    pl = parsed_list
+    JSON::Validator.validate(schema_fname, pl)
```
ptrikutam: This makes me happy. Good stuff.

lib/etl/transform/transform_json.rb
```diff
@@ -0,0 +1,38 @@
...
+  rescue => error
+    ap "Error opening json file: #{error}"
+  end
+
+  private
+
+  attr_reader :input_json, :schema_fname
+  
+  def schema_fname
+    raise NotImplementedError.new("#{self.class.name}#area is an abstract method.")
```
ptrikutam: I'm not sure I understand the error message here. In particular, `#{self.class.name}#area`. Could you clarify?

lib/etl/transform/transform_json.rb
ptrikutam: Rather than assign the variable `pl`, you can actually just use `parsed_list` here and in the line below.

lib/etl/transform/transform_json.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+  
+  def parsed_list
+    JSON.parse(File.read(input_json), :symbolize_names => true)
```
ptrikutam: You can change this method to memoize the return value (so it doesn't do a file read twice) like so:

```
def parsed_list
  @_parsed_list ||= JSON.parse(File.read(input_json), symbolize_names: true)
end
```

That way, in other locations, you can just access `parsed_list` instead of assigning the value to return variable.

Also, in my comment here I've used the more concise way of defining a symbol, changing `:symbolize_names => true` to `symbolize_names: true`.

lib/etl/load/create_new_group_records.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+  end
+
+  private
+
+  attr_reader :attributes_list
+
+  def existing_record(attributes)
+    Group.find_by(name: name(attributes))
```
ptrikutam: I wonder if we should prevent this... Do groups need to necessarily be uniquely named?

lib/etl/load/create_new_group_records.rb
```diff
@@ -0,0 +1,32 @@
+class CreateNewGroupRecords
+  def write(attributes_list)
+    @attributes_list = attributes_list
+
+    attributes_list.each do |attributes|
+      puts "#{attributes["name"]}"
```
ptrikutam: This can actually just be:

```
puts attributes['name']
```

lib/etl/transform/transform_json.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+end
+
+class TransformJsonResources < TransformJson
```
ptrikutam: Should this be singular?

lib/etl/load/create_new_group_records.rb
```diff
@@ -0,0 +1,32 @@
+class CreateNewGroupRecords
+  def write(attributes_list)
```
ptrikutam: Should this class accept an individual set of attributes, rather than a list of attributes? I was under the impression that the ETL script should be loading one record at a time, but maybe I'm mistaken.

lib/etl/extract/process_files_in_s3_bucket_folders.rb
ptrikutam: Taking this a step further, if we change this just to accept one file, perhaps it could actually just be a modification to [`ProcessAllFilesInS3Bucket`](https://github.com/greencommons/commons/blob/master/lib/etl/extract/process_all_files_in_s3_bucket.rb)

What do you think?

ETL.md
```diff
@@ -35,12 +35,20 @@ For more information, see the [`kiba`] `README.md` file.
 
 ## Usage
 
-Then run the rake task to import the records,
+Then run the rake tasks to import the records,
 transform them,
 and load them into the database:
 
 ```
-bundle exec rake etl:import_local_epub
```
gtourtellot: I removed the line because JSON importing can be used for data that originated from EPUBs, so doing this import local epub will be unnecessary.  I'd argue we should just be loading the JSON representation of EPUBs since the EPUB file alone doesn't provide all the info we want -- important metadata, including the book's summary (metadata.abstract field).  That ONIX metadata is already present in the JSON files here: s3://knowledge.commons/valid_data/resources/island_press/

Some extra wins are that the JSON ingestion is faster and the specific implementation of EPUB parsing in ruby chokes on a number of valid files, which doesn't happen for the EPUB parsing to JSON I'm using.

ETL.md
```diff
@@ -51,16 +59,24 @@ kiba ETL source files are in `lib/etl/`
...
 ├── common.rb
-├── convert-islandpress.etl
+├── process_local_json_resources.etl
+├── process_local_json_groups.etl
+├── process_s3_json_resources.etl
+├── process_s3_json_groups.etl
 ├── data
-│   ├── sample_island_press_1.epub
-│   └── sample_island_press_2.epub
+│   ├── good_resource_array.json
```
gtourtellot: Because I think the EPUB parsing as currently is done should be replaced by JSON parsing, I wanted to just reflect that in the ETL.

The reason I didn't actually remove the associated files (sample_island_press_1.epub) in my branch is that I knew we should have this conversation first.

lib/etl/extract/process_files_in_s3_bucket_folders.rb
gtourtellot: It's an array for convenience, to make the .etl files powerful and tidy.

Right now in lib/etl/process_s3_json_resources.etl  and lib/etl/process_s3_json_groups.etl respectively I'm just pulling all resources and groups.  I.e., paths 'valid_data/resources/' and 'valid_data/groups/' can slurp up all the resources all the groups, because of how data is structured in S3.  However, we might want to use just certain folders, and I wanted that to be easy and terse to do.

lib/etl/transform/transform_json.rb
gtourtellot: Thanks :)

gtourtellot: OK, I'll change that.

lib/etl/transform/transform_json.rb
gtourtellot: That's some copy pasta.  I'll fix.

lib/etl/transform/transform_json.rb
gtourtellot: OK, I'll change that.

lib/etl/transform/transform_json.rb
gtourtellot: No.  The JSON schema specifies an array, and hence the processing of the JSON acts on an array of data.

The reason for an array is that in general most of our data is lots of small records.  It would be extremely awkward/ugly to require having a JSON file for each record (of resources, groups, etc.).  As implemented, you can have as many records in a JSON file as you want.  For things like books, having 1 is natural, for things like articles (and our current group data) having a much larger number of resources per JSON file is more natural.

lib/etl/load/create_new_group_records.rb
gtourtellot: KIBA deals with one data file at a time.  The data files we've used thus far are EPUB and JSON.  What's in the data file isn't prescripted by the ETL framework, so one data file can contain multiple records.

lib/etl/load/create_new_group_records.rb
gtourtellot: OK.

lib/etl/load/create_new_group_records.rb
gtourtellot: I didn't want there to be duplicates when running an ingestion script multiple times.  The group name may not be the best key, just as title for resources isn't the best.  Some kind of other unique ID may be the way to go...but that's a larger discussion.

lib/etl/extract/process_files_in_s3_bucket_folders.rb
```diff
@@ -0,0 +1,52 @@
...
+    "#{Rails.root}/tmp"
+  end
+
+  def filtered_files
+    file_keys.grep(/#{file_ext}/)
+  end
+
+  def file_keys
+    arr = []
+    @prefix_folders.each do |prefix_folder|
```
T-Dnzt: Just a quick Ruby tip here, to avoid the extra variable. You can use [`each_with_object`](http://apidock.com/rails/Enumerable/each_with_object) like this:

```
@prefix_folders.each_with_object([]) do |prefix_folder, arr|
  arr += s3.list_objects(bucket_location(prefix_folder)).contents.map(&:key)
end
```

The accumulator is returned by the block so you also don't need to return `arr` anymore. 

(Didn't test the code, might have some typos)

T-Dnzt: Also, since you're using `attr_reader`, you can just use `prefix_folders` (instead of `@prefix_folders`) just like you did for `bucket_name` and `file_ext`.

lib/etl/extract/process_files_in_s3_bucket_folders.rb
```diff
@@ -0,0 +1,52 @@
...
+      arr += s3.list_objects(bucket_location(prefix_folder)).contents.map(&:key)
+    end
+    arr
+  end
+
+  def s3
+    @_s3 = Aws::S3::Client.new
```
T-Dnzt: We might want to add some memoization here? To avoid instantiating a new client for each folder?

```
def s3
  @_s3 ||= Aws.S3::Client.new
end
```

lib/etl/load/create_new_group_records.rb
```diff
@@ -0,0 +1,32 @@
...
+    @attributes_list = attributes_list
+
+    attributes_list.each do |attributes|
+      puts "#{attributes["name"]}"
+      
+      if existing_record attributes
```
T-Dnzt: Maybe we should have parenthesis here, to stay consistent with the rest of the code.

lib/etl/load/create_new_group_records.rb
```diff
@@ -0,0 +1,32 @@
...
+        ap "Name already in database: #{name attributes}" # rubocop:disable Rails/Output
+      else
+        record = Group.create!(attributes)
+
+        ap "Name: #{record.name}" # rubocop:disable Rails/Output
```
T-Dnzt: I'm wondering, do we really need this? I removed them from the other load class because Rubocop wasn't raising any offense anyway. Did you get the offenses without the comments?

lib/etl/load/create_new_group_records.rb
```diff
@@ -0,0 +1,32 @@
+class CreateNewGroupRecords
+  def write(attributes_list)
+    @attributes_list = attributes_list
+
+    attributes_list.each do |attributes|
+      puts "#{attributes["name"]}"
+      
+      if existing_record attributes
+        ap "Name already in database: #{name attributes}" # rubocop:disable Rails/Output
```
T-Dnzt: Parenthesis for `name attributes`?

lib/etl/load/create_new_resource_records.rb
```diff
@@ -0,0 +1,32 @@
+class CreateNewResourceRecords
+  def write(attributes_list)
+    @attributes_list = attributes_list
+
+    attributes_list.each do |attributes|
+      puts "#{attributes["title"]}"
+      
+      if existing_record attributes
+        ap "Title already in database: #{title attributes}" # rubocop:disable Rails/Output
```
T-Dnzt: Parenthesis? And Rubocop comment same as above?

lib/etl/transform/transform_json.rb
T-Dnzt: Yes, very nice! 👏

lib/etl/transform/transform_json.rb
```diff
@@ -0,0 +1,38 @@
...
+  def schema_fname
+    "lib/etl/schema/resource_array_schema.json"
```
T-Dnzt: Single quotes or Rubucop is probably going to complain.

lib/etl/transform/transform_json.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+end
+
+class TransformJsonGroups < TransformJson
+  def schema_fname
+    "lib/etl/schema/group_array_schema.json"
```
T-Dnzt: Same as above - single quotes.

lib/etl/data/bad_group_array.json
```diff
@@ -0,0 +1,17 @@
+[
+	{
```
T-Dnzt: It would be good to have the same indentation as in the other files here (and the files below).

ETL.md
gtourtellot: OK, I agree with you.  I'll update the documentation.

lib/etl/extract/process_files_in_s3_bucket_folders.rb
gtourtellot: OK, good tips, I'll make those changes.

lib/etl/extract/process_files_in_s3_bucket_folders.rb
gtourtellot: OK.

lib/etl/load/create_new_group_records.rb
gtourtellot: OK

lib/etl/load/create_new_group_records.rb
gtourtellot: OK

lib/etl/load/create_new_group_records.rb
gtourtellot: I really don't know.  I can try removing them.

lib/etl/load/create_new_resource_records.rb
gtourtellot: OK

lib/etl/transform/transform_json.rb
gtourtellot: OK, I haven't noticed Rubucop complaining; the hound CI passed.  But if single quotes is more standard, I'll change this.

lib/etl/transform/transform_json.rb
gtourtellot: OK

lib/etl/data/bad_group_array.json
gtourtellot: OK

gtourtellot:
@T-Dnzt, in terms of tests...good point, I agree we should add some tests.  

There is this issue previously added for etl tests, so I was thinking a separate PR could tackle that.  https://github.com/greencommons/commons/issues/58

Curious to know how you all want to set up tests.

T-Dnzt:
After some more thinking, I think we should only pass one record at a time through the various steps. I really don't think we'd lose much efficiency by moving the loops that are currently in the transform and load steps back into the extract step. 

The extract step simply needs to check if the received data is an array or not, and loop through it before passing each record to the next step. It keeps the ETL logic unified and make it easier to chain or reorganize them in the future.

@gtourtellot I'm happy to discuss more about this to see if we can agree on something.

About the tests, I think we can have a test running the same code found in the ETL rake task and check if the resources/groups were created as expected after that.

The tests weren't added in the previous ETL PR and we still don't have them :) That's why I believe any PR should have at least a few working tests - we can always expand on that later. Adding one test to run the steps and ensure that a resource was created shouldn't take too long, the sample epub/json that you added can actually be used (we should probably move them into `/spec/support/samples` or `/spec/samples`). I could actually work on it if you'd like me to.

cc: @ptrikutam

gtourtellot:
On Mon, Jan 23, 2017 at 8:33 AM, Thibault <notifications@github.com> wrote:

> After some more thinking, I think we should only pass one record at a time
> through the various steps. I really don't think we'd lose much efficiency
> by moving the loops that are currently in the transform and load steps back
> into the extract step.
>
> The extract step simply needs to check if the received data is an array or
> not, and loop through it before passing each record to the next step. It
> keeps the ETL logic unified and make it easier to chain or reorganize them
> in the future.
>
> @gtourtellot <https://github.com/gtourtellot> I'm happy to discuss more
> about this to see if we can agree on something.
>

I was thinking similar thoughts over the weekend.  I think we're in
agreement, since what you suggest satisfies my desire to allow the JSON
files to contain arrays, and also allows more modularity of record
processing.

There's also a little bit of other modularization/cleanup that can be done
at the same time.  I.e., extract/process_all_files_in_s3_bucket.rb has
hardcoded paths and file types, which is unnecessary/undesirable.

About the tests, I think we can have a test running the same code found in
> the ETL rake task and check if the resources/groups were created as
> expected after that.
>
> The tests weren't added in the previous ETL PR and we still don't have
> them :) That's why I believe any PR should have at least a few working
> tests - we can always expand on that later. Adding one test to run the
> steps and ensure that a resource was created shouldn't take too long, the
> sample epub/json that you added can actually be used (we should probably
> move them into /spec/support/samples or /spec/samples). I could actually
> work on it if you'd like me to.
>
> cc: @ptrikutam <https://github.com/ptrikutam>
>
I'd be in favor of turning over the entire PR to you, or as much of it as
you and Pavan are ok happy with.  Arguably faster/easier for you to make
the changes in a sound ruby way.

T-Dnzt:
@gtourtellot Okay, I will be taking over the PR and request your review once I made the changes ;)

gtourtellot:
Great, thanks.

Issue #99, opened 2017-01-09
Resource View Page


ptrikutam:
Wireframes: https://invis.io/BH8GKH38D#/185307862_RESOURCE_-_View

We'd like to build out the base UI for the Resource page.

Most of it should be straightforward, but for sake of specificity I'd like to clarify a few things:

- For the time being, please ignore (don't display) the "Discuss" section
- Please skip the "Content" section for now.
- Try to extract the actual metadata for the Resource in the "About" section -- currently we're pulling Title, Creators, Date (publish date), Publisher, and a "Metadata" field. Can you investigate what type of data is included in the "Metadata" field and see if we can extract that to be displayed in the Resource view?
- For the "Explore" section, let's go for things that share the same tags to start. Ideally, this should use a similar mechanism as the one described in https://github.com/greencommons/commons/issues/98 in the "You May Also Like" section. Basically, we want a centralized place to recommend similar / related (directly or tangentially) items for a given resource, list, group, etc. Happy to talk through this part.

T-Dnzt:
Maybe we can close this? :)

PR #110, opened 2017-01-22
Resource page design

T-Dnzt:
Related Issue: #99 

This pull request contains the new design for the resource page. It is not the final page, since some sections are empty or act as placeholders (tags, content, hidden discuss, temporary explore section).

The suggestion mechanism will be implemented in a future Pull Request, therefore we will need to update the `show` action in the `ResourcesController` to use it. The group view page was also modified in order to reuse the same styles/components in the resource page.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -2,6 +2,8 @@ class ResourcesController < ApplicationController
   before_action :set_resource, only: [:show]
 
   def show
+    # TODO: Switch this to suggestion mechanism once implemented
+    @suggestions = Group.limit(6)
   end
 
   private
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -1,0 +1,13 @@
+class ResourcesController < ApplicationController
+  before_action :set_resource, only: [:show]
+
+  def show
+  end
+
+  private
+
+  def set_resource
+    @resource = Resource.find(params[:id])
+    authorize @resource
+  end
+end
--- a/app/policies/resource_policy.rb
+++ b/app/policies/resource_policy.rb
@@ -1,0 +1,12 @@
+class ResourcePolicy < ApplicationPolicy
+  attr_reader :user, :resource
+
+  def initialize(user, resource)
+    @user = user
+    @resource = resource
+  end
+
+  def show?
+    true
+  end
+end
--- a/app/views/groups/show.html.slim
+++ b/app/views/groups/show.html.slim
@@ -105,24 +105,4 @@
             
   .row.page-details__row
     .col-lg-12
-      .row.page-details__row--half
-        .col-lg-12
-          h3 DISCUSS
-      .row
-        .col-lg-12
-          = render 'comments/comment', resource: nil
-          = render 'comments/comment', resource: nil
-          = render 'comments/comment', resource: nil
-          = render 'comments/subcomment', resource: nil
-          = render 'comments/comment', resource: nil
-      .row.page-details__row
-        .col-lg-12.text-center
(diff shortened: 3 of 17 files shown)


app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,15 @@
+class ResourcesController < ApplicationController
+  before_action :set_resource, only: [:show]
+
+  def show
+    # TODO: Switch this to suggestion mechanism once implemented
+    @suggestions = Group.limit(6)
```
ptrikutam: This may also propose different items, like other Resources, Lists, People. Is that what you're envisioning here? (No change needed for this PR, but just wanted to make sure that was clear)

app/policies/resource_policy.rb
```diff
@@ -0,0 +1,12 @@
+class ResourcePolicy < ApplicationPolicy
```
ptrikutam: I presume this class will have edit and other types of permissions once we develop it further, yes?

app/views/groups/show.html.slim
```diff
@@ -25,39 +25,22 @@
...
-                    span.glyphicon.glyphicon-plus aria-hidden='true'
-        .col-lg-9
-          p.group-details__tag-list
-            span.glyphicon.glyphicon-tag.glyphicon--right
-            a href="/" tag
-            | , 
-            a href="/" another tag
-            | , 
-            a href="/" solar panels
+      = render 'shared/components/tags'
```
ptrikutam: 🙌

app/views/groups/show.html.slim
```diff
@@ -120,26 +103,6 @@
...
-          = link_to 'LOAD MORE COMMENTS', '#', class: 'btn btn-primary'
-          
-      .row
-        .col-lg-12
-          = form_tag '/', class: 'comment-form' do
-            .form-group
-              = text_area_tag :comment, '', class: 'form-control comment-form__textarea', placeholder: 'Add a comment...'
-            .form-group.pull-right
-              = button_tag 'SUBMIT', type: 'submit', class: 'form-control btn btn-primary'
+      = render 'shared/components/discuss'
```
ptrikutam: Excellent. This may be replaced with Disqus soon, so I'm glad it's in one place.

app/views/resources/show.html.slim
```diff
@@ -0,0 +1,73 @@
+.page-details
+  .row.page-details__row
+    .col-xs-6.col-md-4
+      h4
```
ptrikutam: Can we make this `h6` instead? `h4` is a tad large.

app/views/resources/show.html.slim
```diff
@@ -0,0 +1,73 @@
...
+        span class="glyphicon glyphicon--right glyphicon-#{resource_icon(@resource)}"
+        = @resource.resource_type.capitalize
+    
+    .col-xs-6.col-md-8
+      .resource-actions
+        = render 'shared/components/btn_count', path: '#',
```
ptrikutam: Small nit: on a mobile screen, this button gets messed up: https://cl.ly/1F3H2N0R2X15

app/controllers/resources_controller.rb
T-Dnzt: Yes, not sure how I'm going to do it yet, but that's the plan. :)

app/policies/resource_policy.rb
T-Dnzt: Yes, totally, just wanted to start with the bare minimum since I don't have any design or anything for the other actions.

app/views/resources/show.html.slim
T-Dnzt: Changed the columns, now it works except for screens smaller than 290px. If that's not good enough, I will fix the design in another PR. ;)

Issue #108, opened 2017-01-15
Use update_document method to re-index updated record 


T-Dnzt:
I think we should separate the `create`/`update` callback [here](https://github.com/greencommons/commons/blob/master/app/models/resource.rb#L26) and start using the `update_document` from [there](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L389). I'm worried about duplicates and was able to produce a bug by updating a record, searching for its old name and still finding it even though it should have been de-indexed.

It would also be interesting to unify the way we interact with ES by not using the `ElasticSearch::Modal.client` directly and instead relying on the methods provided by [elasticsearch-rails](https://github.com/elastic/elasticsearch-rails).

[delete_document](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L357) could replace the three lines below for example.

```
  def remove
    if search_index_callbacks_enabled?
      Elasticsearch::Model.client.delete(
        index: model_name.constantize.index_name,
        type: model_name.downcase,
        id: id,
      )
    else
      log_callback_warning
    end
  end
```

@ptrikutam Let me know if that's something you'd like me to work on or not.

ptrikutam:
I like this idea. One question: does the `update_document` method create a record in the index if it doesn't exist already? Judging from [this line](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L404) it seems like it does, but I wasn't sure.

How long do you think this will take you?

T-Dnzt:
Yes, based on the code, it should index the record if it doesn't exist yet. I can run some tests before implementing it to double-check.

For the timeframe, since we already have the setup to create/delete from the index, it won't take more than an hour (probably much less).

ptrikutam:
Got it. OK, yes please go ahead and implement this. Thanks!

T-Dnzt:
The implementation ended up looking different from what I originally wanted to do, but the goal has been accomplished. I learned a lot about ES and the way we interact with it so that's a great point for the rest of the project. Closing this issue!

ptrikutam:
Great, thanks.​

Issue #44, opened 2016-12-01
Searching for records either via the controller or model should return AR objects, not ES objects


ptrikutam:

adarsh:
From the [`elasticsearch-model`docs](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#search-results-as-database-records):

#### Search results as database records

Instead of returning documents from Elasticsearch, the `records` method will return a collection
of model instances, fetched from the primary database, ordered by score:

```ruby
response.records.to_a
# Article Load (0.3ms)  SELECT "articles".* FROM "articles" WHERE "articles"."id" IN (1, 2)
# => [#<Article id: 1, title: "Quick brown fox">, #<Article id: 2, title: "Fast black dogs">]
```

The returned object is the genuine collection of model instances returned by your database,
i.e. `ActiveRecord::Relation` for ActiveRecord, or `Mongoid::Criteria` in case of MongoDB.

This allows you to chain other methods on top of search results, as you would normally do:

```ruby
response.records.where(title: 'Quick brown fox').to_a
# Article Load (0.2ms)  SELECT "articles".* FROM "articles" WHERE "articles"."id" IN (1, 2) AND "articles"."title" = 'Quick brown fox'
# => [#<Article id: 1, title: "Quick brown fox">]

response.records.records.class
# => ActiveRecord::Relation::ActiveRecord_Relation_Article
```

The ordering of the records by score will be preserved, unless you explicitly specify a different
order in your model query language:

```ruby
response.records.order(:title).to_a
# Article Load (0.2ms)  SELECT "articles".* FROM "articles" WHERE "articles"."id" IN (1, 2) ORDER BY "articles".title ASC
# => [#<Article id: 2, title: "Fast black dogs">, #<Article id: 1, title: "Quick brown fox">]
```

The `records` method returns the real instances of your model, which is useful when you want to access your
model methods -- at the expense of slowing down your application, of course.
In most cases, working with `results` coming from Elasticsearch is sufficient, and much faster. See the
[`elasticsearch-rails`](https://github.com/elasticsearch/elasticsearch-rails/tree/master/elasticsearch-rails)
library for more information about compatibility with the Ruby on Rails framework.

When you want to access both the database `records` and search `results`, use the `each_with_hit`
(or `map_with_hit`) iterator:

```ruby
response.records.each_with_hit { |record, hit| puts "* #{record.title}: #{hit._score}" }
# * Quick brown fox: 0.02250402
[...]

PR #111, opened 2017-01-22
Implement ES update mechanism

T-Dnzt:
Related Issues: #108, #43 and #44 

This pull request adds a new indexing process for existing records. Instead of re-indexing entire documents every time an update is made, it will select the changed fields and send them Elastic Search. In case the record wasn't indexed for some reason, it will perform a full indexing.

I also made the remove process more robust by failing silently if the record wasn't indexed (using the same mechanism as the update: by rescuing from `Elasticsearch::Transport::Transport::Errors::NotFound`.

### DIFF
--- a/app/service/search_index.rb
+++ b/app/service/search_index.rb
@@ -20,8 +20,6 @@ class SearchIndex
   end
 
   def update
-    return add unless changed.any?
-
     perform do
       data = elasticsearch_hash.merge(body: { doc: changes })
       Elasticsearch::Model.client.update(data)
--- a/app/service/search_index.rb
+++ b/app/service/search_index.rb
@@ -9,35 +9,57 @@ class SearchIndex
     end
   end
 
-  def initialize(model_name:, id:)
+  def initialize(model_name:, id:, changed: nil)
     @model_name = model_name
     @id = id
+    @changed = changed
   end
 
   def add
-    if search_index_callbacks_enabled?
-      record = model_name.constantize.find(id)
-      record.__elasticsearch__.index_document
-    else
-      log_callback_warning
+    perform { record.__elasticsearch__.index_document }
+  end
+
+  def update
+    return add unless changed.any?
+
+    perform do
+      data = elasticsearch_hash.merge(body: { doc: changes })
+      Elasticsearch::Model.client.update(data)
     end
+  rescue Elasticsearch::Transport::Transport::Errors::NotFound
+    add
   end
 
   def remove
-    if search_index_callbacks_enabled?
-      Elasticsearch::Model.client.delete(
-        index: model_name.constantize.index_name,
-        type: model_name.downcase,
-        id: id,
-      )
-    else
-      log_callback_warning
-    end
+    perform { Elasticsearch::Model.client.delete(elasticsearch_hash) }
+  rescue Elasticsearch::Transport::Transport::Errors::NotFound
   end
 
   private
 
-  attr_reader :id, :model_name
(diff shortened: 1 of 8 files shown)


app/service/search_index.rb
```diff
@@ -9,35 +9,57 @@ def self.log_elasticsearch_warning(message)
...
-    if search_index_callbacks_enabled?
-      record = model_name.constantize.find(id)
-      record.__elasticsearch__.index_document
-    else
-      log_callback_warning
+    perform { record.__elasticsearch__.index_document }
+  end
+
+  def update
+    return add unless changed.any?
```
ptrikutam: This will add a record if there's no changes, correct? Does this only occur when a new record is created?

app/service/search_index.rb
```diff
@@ -9,35 +9,57 @@ def self.log_elasticsearch_warning(message)
...
+
+    perform do
+      data = elasticsearch_hash.merge(body: { doc: changes })
+      Elasticsearch::Model.client.update(data)
```
ptrikutam: What's the motivation to use this method of updating instead of using the method [described in the `elasticsearch-model` docs](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model)?

```
record.__elasticsearch__.update_document
```

app/service/search_index.rb
ptrikutam: Rather, better phrased: when does the situation occur when there's no changes passed to the `update` method?

T-Dnzt:
Thanks for checking. Since I had some issues while writing the code for this PR, I'm going to double-check the stuff that you mentioned - just want to be sure everything is working as expected.

app/service/search_index.rb
T-Dnzt: Hmm, I think you're right. I had a lot of issues with the different callbacks / changes methods (which are different if you use `after_commit`, `after_update`, etc) and I might have forgotten to remove this. Gonna run a few more tests and see if I can remove it.

app/service/search_index.rb
T-Dnzt: Mostly because it wasn't working as expected. Since we're using Sidekiq and we can't pass Ruby objects through it easily (that's why we only pass the `id` and `class`), I wasn't always getting the correct data in ElasticSearch. Going to double-check this and see if I can switch back now that I'm using `after_commit` vs `after_update`.

T-Dnzt: Okay, I remember now. It wasn't working because we use Sidekiq and load the model from the database in the `SearchIndex` instead of using the loaded model on which the changes were made, therefore it doesn't have the changes list available for `update_document`. That's why instead I use the ES Client and pass it the changes directly through Sidekiq.

app/service/search_index.rb
T-Dnzt: Going to remove this since I can't think of a situation where it would happen.

T-Dnzt:
> I deleted an item from the index, then called the UpdateIndexJob on it without passing any changes, and from what I could tell it didn't do anything. I'm sure I'm using it wrong, so I don't want to hold up the PR any more because of that.

I just did that test myself, with some changes, and it worked properly. Your approach seems fine though, that's what I just did manually and what happens in this [test too](https://github.com/greencommons/commons/pull/111/files#diff-a41680acf76829cefd4b88af66a918adR51). 

I'm going to get this merged but let's keep an eye open for this stuff. I will be running some tests in the live application when searching for groups (since we have the pages to update them) and might open some issue in the future if I notice anything wrong.

PR #112, opened 2017-01-22
Implement basic search page

T-Dnzt:
Related issue: #98 (Points 1 and 2)

This pull request adds a simple search result page, stripped down from any complexity that will be added in future pull requests. There are many small changes on the front-end side, which made sense to make this PR stands on its own.

The ElasticSearch score was added for each record, I don't think this will be the final value being shown but it's a start. There is no filtering, no sorting and no pagination for now.

I'm also wondering if we should add [Font Awesome](http://fontawesome.io/) to the project to replace the very limited amount of icons coming with Bootstrap.

## Screenshots

screen shot 2560-01-22 at 11 53 44 pm

### DIFF
--- a/app/assets/stylesheets/components/_top-page-form.scss
+++ b/app/assets/stylesheets/components/_top-page-form.scss
@@ -1,3 +1,3 @@
-.add-member-form {
+.top-page-form {
   margin-top: 36px;
 }
--- a/app/helpers/summary_cards_helper.rb
+++ b/app/helpers/summary_cards_helper.rb
@@ -1,10 +1,10 @@
 module SummaryCardsHelper
   def summary_card_for(resource)
     {
-      Group => 'shared/summary_cards/group',
-      GroupsUser => 'shared/summary_cards/member',
-      List => 'shared/summary_cards/list',
-      Resource => 'shared/summary_cards/resource',
-    }[resource.class]
+      'Group' => 'shared/summary_cards/group',
+      'GroupsUser' => 'shared/summary_cards/member',
+      'List' => 'shared/summary_cards/list',
+      'Resource' => 'shared/summary_cards/resource',
+    }[resource.class.to_s]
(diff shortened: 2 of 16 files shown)


app/assets/stylesheets/components/_top-page-form.scss
```diff
@@ -1,3 +1,3 @@
-.add-member-form {
+.top-page-form {
```
T-Dnzt: If it's not, let's invent it!

app/helpers/summary_cards_helper.rb
```diff
@@ -1,10 +1,10 @@
 module SummaryCardsHelper
   def summary_card_for(resource)
     {
-      Group => 'shared/summary_cards/group',
-      GroupsUser => 'shared/summary_cards/member',
-      List => 'shared/summary_cards/list',
-      Resource => 'shared/summary_cards/resource',
-    }[resource.class]
+      'Group' => 'shared/summary_cards/group',
```
T-Dnzt: Weirdest thing ever. It wasn't working with symbols for the `Resource` class so I decided to switch to strings. I don't know why it happened (only for `Resource`) and didn't want to spend too much time trying to figure it out. I did some checking of course, and was getting the correct `Resource` class (being passed and in the hash) but for some reason they did not match.

spec/rails_helper.rb
```diff
@@ -38,7 +40,9 @@
   end
 
   def reset_resource_index
```
T-Dnzt: Yes, good catch!

T-Dnzt:
Just added [this feature test](https://github.com/greencommons/commons/pull/112/commits/a97063be48eec9f3750688a6727906b2cc9249e5#diff-a41680acf76829cefd4b88af66a918adR75) to check that we were getting resources, groups and lists.

PR #113, opened 2017-01-24
Support for JSON and ETL system refactoring

T-Dnzt:
Based on [PR #109](https://github.com/greencommons/commons/pull/109).

Origin description:

```
ETL.md shows rakes to use and new ETL files.
Groups and Resources JSON schema are compatible with and extend db/schema.rb defs.
Local data in lib/etl/data are useful for simple tests.
Ingestion of JSON from S3 is supported.
S3 JSON data derives from sources including EPUBs augmented with ONIX data.
```

What changed from the previous PR:

- `ETL.md` update to show all the ETL pipelines that are available, and which ones should be used.
- Use of the same pattern for all ETL filenames.
- Modify the sources to be able to receive either one folder or an array of folders to parse, making it easier to parse multiple folders from S3 in one process (originally added by @gtourtellot).
- Modify the destinations to only handle one record at a time. Also, implemented a parent class from which the resource and group classes inherit.
- Creation of normalizers ([source](http://thibautbarrere.com/2015/06/25/how-to-explode-multivalued-attributes-with-kiba/)): Since a transformer class cannot yield more than one record, I had to remove the `TransformJson` class. Instead, a normalizer takes care of splitting the array of records and passing each one of them to the destination.
- Removal of all the very similar Rake tasks running the pipelines. Instead, the ETL class takes care of the logic and some metaprogramming generates the rake task based on the list of supported ETLs.
- Addition of tests for the local pipelines (easy to test). Testing the S3 ETL pipelines requires more work and probably some discussion about the best way to stub S3. Therefore, S3 pipelines are not tested in this PR.

__Some of the JSON files are showing extra indentation on GitHub but not in local. I'm not sure where this is coming from, but if you have any idea, let me know.__


georgeatmuseami:
All the changes look very good, and the rakes run correctly for me.  Nice job!

Wrt your comment about the JSON file indentation looking differently locally vs in github, I don't notice any difference.

Wrt testing with s3...we could create a folder on s3 with a small set of dummy files.  That folder could be made public so that anyone could have the test pass.  If you want to bring back this branch before we tackle that, though, that'd be fine with me; see what @ptrikutam thinks.

gtourtellot:
One followup:  although the ETL rakes did run for me correctly as I
mentioned before, I see 19 offenses when I run:  bundle exec rake

T-Dnzt:
@gtourtellot Thanks for letting me know about this, just made this [PR](https://github.com/greencommons/commons/pull/114) to fix the problem.

georgeatmuseami:
LGTM 👍

PR #114, opened 2017-01-25
Change Rubocop rake task to use .rubocop config

T-Dnzt:
Currently, when running `bundle exec rake` (which runs `rspec` and the `rubocop` rake task), we are getting offenses that should be ignored by the configuration defined in [.rubocop.yml](https://github.com/greencommons/commons/blob/fix-rubocop-rake-task/.rubocop.yml).

Changing this rake task to use `RuboCop::RakeTask.new` seems to fix the issue.


georgeatmuseami:
That seems to have fixed it.  LGTM 👍

PR #117, opened 2017-02-01
Search Page Filtering/Sorting/Pagination

T-Dnzt:
Related Issue: #98 

This pull request comes with a complete filtering & sorting system for the search page. Pagination is also included. It's quite big but it's all about showing the most relevant results to the user based on his query/filters/sorting options.

Since we are searching through ElasticSearch and not a simple PG database, it was necessary to create custom classes that build up a hash containing the search parameters that will be passed to the ElasticSearch client.

## The Search Builders

- [`SearchBuilders::Builder`](https://github.com/greencommons/commons/blob/87d911268e7df96e1f618075851eea3f076bdd6c/app/services/search_builders/builder.rb)
- [`SearchBuilders::Query`](https://github.com/greencommons/commons/blob/5321947288a4535b0abb4d112d47dab0d30a9c58/app/services/search_builders/query.rb)
- [`SearchBuilders::ModelLister`](https://github.com/greencommons/commons/blob/87d911268e7df96e1f618075851eea3f076bdd6c/app/services/search_builders/model_lister.rb)
- [`SearchBuilders::ResourceTypeFilter`](https://github.com/greencommons/commons/blob/5321947288a4535b0abb4d112d47dab0d30a9c58/app/services/search_builders/resource_type_filter.rb)
- [`SearchBuilders::Sorter`](https://github.com/greencommons/commons/blob/5321947288a4535b0abb4d112d47dab0d30a9c58/app/services/search_builders/sorter.rb)

These classes are used in the `SearchController` to build the appropriate ES search parameters:

```
builder = SearchBuilders::Builder.new(
  query: @query,
  filters: @filters,
  sort: @sort,
  dir: @dir
).search.filter_by_resource_type.sort

@results = Elasticsearch::Model.search(*builder.to_elasticsearch).
           page(params[:page] || 1).per(10)
```

Each class is specialized and takes care of modifying a few things in the basic ES search options initiated in the `Builder` class.

## Pagination with Kaminari

Kaminari is used for pagination, and hooked into ElasticSearch Rails in the `initializers/kaminari.rb` file. This leads to the `page` and `per` parameters being added to the ES query (instead of doing it in the PG database).

## Updating Circle.yml to use a specific version of ES

I had a few failures on CircleCI because they were only providing an older version of ElasticSearch (1.7.something) and I built the search options using the documentation of the production version (2+). Once changed to install a specific version based on some code provided in the CircleCI documentation, everything started working again.

## Coming Soon

The following wasn't added to the PR because it was already big enough. :)

Currently, the filters don't include date ranges. That's because we need to have a chat about it. I wasn't able to find a way to sort using either the `metadata.date` or the `created_at` date. I'm recommending the creation of a new field in the `Resource`, `Group` and `List` class named that will be used to sort them on the search page. It will probably be the publication date for the resources, the creation date for the groups, lists and users and should be automatically updated with an `after_*` callback.

Once we have this, we can update the `SearchBuilders::Sorter` class to use it (just have to change one value). There's also no way to specify which sorting direction the user wants in the front-end but the logic is there in the backend. Let me know if you have any requirements for it and I will fix it.

Finally, I didn't touch the score but I'm thinking about making it a percent (based on the score provided by ES of course) and put some nice CSS around it to make it look good. 

## Screenshots

screen shot 2560-02-01 at 5 10 44 pm


app/views/kaminari/_gap.html.slim
```diff
@@ -0,0 +1,2 @@
+li.disabled
+  = link_to raw(t 'views.pagination.truncate'), '#'
```
T-Dnzt: To be honest, I don't really know, these are the views generated by Kaminari with the Bootstrap style and the `slim` format. Didn't really mess with them much :D

app/assets/javascripts/checkbox.js
```diff
@@ -0,0 +1,22 @@
+$(document).on('ready turbolinks:load', function () {
```
T-Dnzt: Yeah, I also wanted to get rid of the FILTER / SORT buttons with React and just let stuff happen automatically anytime a user click/change something. But the current implementation will be easier for the server to handle more traffic (way less requests with buttons).

circle.yml
```diff
@@ -5,12 +5,17 @@ machine:
     version:
       2.3.2
 
-  services:
-    - elasticsearch
-
   environment:
     CIRCLE_ENV: test
 
+dependencies:
```
T-Dnzt: Not sure what you mean by "brittle configuration" and if you're actually asking me. I guess we reached the limits of my English skills! :p

But yeah, when I saw the tests were doing stuff not supported by ES in CircleCI, I decided to check the versions.

app/views/search/show.html.slim
```diff
@@ -1,35 +1,65 @@
...
+= form_tag search_path, method: :get do
+  .row.page-details__row--half
+    .col-xs-12.col-sm-6.page-details__row
+      h4= "#{@results.count} Results"
+    .col-xs-12.col-sm-6
+      .row
+          .col-xs-8.col-sm-4.col-sm-offset-4
+            = hidden_field_tag :query, @query
+            = hidden_field_tag :dir, @dir || 'desc'
+            = select_tag :sort, options_for_select([ 'SCORE', 'DATE' ], @sort || 'SCORE'), class: 'form-control'
```
T-Dnzt: Yes, totally, I will update it.

PR #118, opened 2017-02-01
Implement Tagging system

T-Dnzt:
This pull request implements the tagging system for groups and resources. It uses the `act-as-taggable-on` gem to do so and comes with a React component that can be used to add tags to any model through the `Tags` controller.

I tried to add a feature spec but wasn't able to get the button the React component clicked and do its thing... it kept failing so I'm submitting the PR without tests and will spend some time trying to figure out the issue and create a separate pull request.

It could also be interested to look into [jasmine-react-rails](https://github.com/onfido/jasmine-react-rails) to test the components directly.


app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,21 @@
+class TagsController < ApplicationController
```
T-Dnzt: While making this, I was thinking about taking the `Api::AutocompleteController` out of the `Api` namespace since these controllers are not part of an API for external developers but simply internal controllers returning `JSON` documents.

app/policies/resource_policy.rb
```diff
@@ -9,4 +9,8 @@ def initialize(user, resource)
   def show?
     true
   end
+
+  def update?
+    true
```
T-Dnzt: This should be changed, but I don't know who can update resources?

app/views/resources/show.html.slim
```diff
@@ -20,7 +20,10 @@
 
   .row.page-details__row
     .col-xs-12
-      = render 'shared/components/tags'
+      = react_component 'Tags', { authenticity_token: form_authenticity_token,
+                                  submit_url: resource_tags_path(@resource),
+                                  tags: @resource.tags,
+                                  can_create: true }, { prerender: true }
```
T-Dnzt: The `can_create` here should also be changed.

app/models/group.rb
```diff
@@ -1,5 +1,6 @@
 class Group < ApplicationRecord
   include Indexable
+  acts_as_taggable
```
ptrikutam: Do we need this? I see it already on line 6.

app/views/resources/show.html.slim
ptrikutam: Agreed. For the time being I think we can allow free tagging by anyone.

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  TAGGABLE_MODELS = %w(Group Resource).freeze
```
ptrikutam: We'd also like Lists to be taggable. Should we add that in a later PR once the List models are all set up?

app/controllers/tags_controller.rb
ptrikutam: Sounds good to me.

app/policies/resource_policy.rb
ptrikutam: Hmm... I would think the only person who can edit a Resource is its owner.

app/views/shared/components/_tags_list.html.slim
```diff
@@ -0,0 +1,6 @@
+p
+  span.glyphicon.glyphicon-tag.glyphicon--right
+  - tags.each_with_index do |tag, i|
+    = link_to tag.name, root_path
```
ptrikutam: Eventually we're going to want to have this link to the search path instead of the `root_path`. I presume once you've set up the tagging <> ES stuff we can add that in.

app/controllers/tags_controller.rb
T-Dnzt: Yes, once this PR and #120 are merged, it will be super easy to make lists taggable. I can do it when we implement the pages to navigate lists.

app/views/shared/components/_tags_list.html.slim
T-Dnzt: Oh, I thought we would have separate pages for tags, which is why I left it like this, but linking it to the search is fine as well.

app/policies/resource_policy.rb
T-Dnzt: But the owner is optional for resources (we mostly have orphans resources actually).

ptrikutam: Yeah, for the time being this is fine. We don't really have a way for people to edit resources on the site now anyway.

app/controllers/tags_controller.rb
ptrikutam: 👍

PR #120, opened 2017-02-02
Suggestions for search and resource based on tags

T-Dnzt:
Issue: #98 & #99 
Depends on PR #118 for tags and PR #117 for the correct version of ES in CircleCI (hence the failing tests).

# To do

- [x] Add feature tests
- [x] Ensure that the `Suggester::Tags` works correctly on the search page

# Explore / You may also like Section

This pull request adds the __Explore__ section to resources and the __You may also like__ section containing up to 6 suggestions based on the records tags.

I had quite a few issues with ES indexing since the tags are store in external tables. At first, I tried using the caching system from `acts-as-taggable-on` but it was so damn buggy that I switch to a custom and simple one. 

# Changes

- All the tags of the searched records are now used to find related records
- Clicking on a tag will take you to the search page with that tag as the query
- Feature specs for related records on search and resource pages
- Use of `cached_tags` everywhere to speed things up and prevent tons of queries

### DIFF
--- a/app/models/concerns/taggable.rb
+++ b/app/models/concerns/taggable.rb
@@ -1,0 +1,17 @@
+module Taggable
+  extend ActiveSupport::Concern
+
+  included do
+    acts_as_taggable
+
+    after_touch do
+      update(cached_tags: tag_list)
+    end
+  end
+
+  def as_indexed_json(_options = {})
+    json = as_json
+    json['tags'] = json.delete('cached_tags')
+    json
+  end
+end
--- a/config/initializers/acts_as_taggable_on.rb
+++ b/config/initializers/acts_as_taggable_on.rb
@@ -1,2 +1,3 @@
 ActsAsTaggableOn.remove_unused_tags = true
 ActsAsTaggableOn.force_lowercase = true
+ActsAsTaggableOn::Tagging.belongs_to :taggable, polymorphic: true, touch: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,11 @@ AllCops:
 Metrics/LineLength:
   Max: 100
   Enabled: true
+Metrics/BlockLength:
+  Exclude:
+      - 'config/routes.rb'
+      - 'spec/**/*.rb'
+      - '**/*.rake'
 Style/FrozenStringLiteralComment:
   Enabled: false
   EnforcedStyle: when_needed
@@ -27,11 +32,6 @@ Style/FrozenStringLiteralComment:
     - always
 Style/MultilineOperationIndentation:
   Enabled: false
-Style/BlockLength:
-  Exclude:
-      - 'config/routes.rb'
-      - 'spec/**/*.rb'
-      - '**/*.rake'
 Style/StringLiterals:
   EnforcedStyle: single_quotes
   SupportedStyles:
--- a/app/assets/javascripts/components/tags.jsx
+++ b/app/assets/javascripts/components/tags.jsx
@@ -75,8 +75,8 @@ var Tags = React.createClass({
             <span className="glyphicon glyphicon-tag glyphicon--right"></span>
             {this.state.tags.map(function(tag, i) {
               return (
-                <span key={tag.name}>
-                  <a href="/">{tag.name}</a>
+                <span key={tag}>
+                  <a href={'/search?query=' + tag}>{tag}</a>
                   {i == _this.state.tags.length - 1 ? '' : ', '}
                 </span>
               )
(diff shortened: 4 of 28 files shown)


T-Dnzt:
Sure, I will answer your questions. For now, the way I tested it was to add a tag to one resource, and the same tag to a group. After refreshing the resource page, the group should show in the list of the EXPLORE section. Just realized I haven't tested it on the search page and will do now.

And this PR could use some actual feature specs too. Will add that.

app/models/concerns/taggable.rb
```diff
@@ -0,0 +1,17 @@
+module Taggable
```
T-Dnzt: Yes, so the `cached_tags` is one table field defined for the `Group` and `Resource` models. The problem with `acts-as-taggable-on` is that the tags are saved after the record has been saved, therefore when we send the model for ES indexing, the tags are not available yet.

The only way I found to do this was like this:

1. New tag is added to record. Record is saved and goes for indexing (but nothing will be updated because no fields has changed, yet).
2. acts-as-taggable-on kicks in, saving the tag and linking it to the record.
3. Because `touch: true` is defined on the relation between the taggings and the records (that's the line you ask me to explain below), the tagging record will touch (updating the `updated_at` basically) the record.
4. The `after_touch` callback is triggered, the list of tags stored in the record is updated by querying for the tag list from the `Tags`/`Taggings` tables.
5. The `after_commit on [:update]` is finally triggered, replaces the `cached_tags` by `tags` in the `as_indexed_json` (because that's how we want it to be called in the ES index) and send it for ES re-indexing/update.

Oh and the `as_indexed_json` is the hash that will be turn into a JSON and sent to ES for indexing, here I just override it to replace the `cached_tags` key by `tags`. We use it in the `SearchIndex` class as follow:

```
def changes
    record.__elasticsearch__.as_indexed_json.select do |k, _|
      changed.include?(k.to_s)
    end
  end
```

Let me know if anything is unclear ;)

config/initializers/acts_as_taggable_on.rb
```diff
@@ -1,2 +1,3 @@
 ActsAsTaggableOn.remove_unused_tags = true
 ActsAsTaggableOn.force_lowercase = true
+ActsAsTaggableOn::Tagging.belongs_to :taggable, polymorphic: true, touch: true
```
T-Dnzt: By default I think the line is (where `taggable` is the group or resource record): 

```
ActsAsTaggableOn::Tagging.belongs_to :taggable, polymorphic: true
```

Here, I just override it to add the `touch: true` that will triggers the related record update.

app/models/concerns/taggable.rb
ptrikutam: Gotcha. So the path goes:

Tag added --> Record Saved --> after_commit triggers indexing --> Tags "touches" Record --> Record sets `cached_tags` from Tags table --> after_commit triggers indexing --> as_indexed_json overrides `tags` field and removes `cached_tags`.

Is that right? I might be oversimplifying this, but would it be possible to skip the `cached_tags` entirely and instead just use `tag_list` in the [`as_indexed_json`](https://github.com/greencommons/commons/pull/120/files#diff-13edb83250a6faff4a1518a8dee4f9a2R14) override? Either way the record gets indexed twice, right? Maybe I'm thinking of the order of operations incorrectly...

Seems pretty complicated, but ES forces us to go this way I suppose. Thanks for working through this.

config/initializers/acts_as_taggable_on.rb
ptrikutam: Got it. Thanks.

ptrikutam:
@T-Dnzt let me know if this is ready for a full re-review.

app/models/concerns/taggable.rb
T-Dnzt: Ah yes, we can probably get rid of the `cached_tags`. I didn't think about removing it because it was needed to have ES work with acts-as-taggable-on caching, but since I'm not using it, I can probably ditch it.

It gets indexed twice, but only the stuff that has changed. The first time, it might be just the title, but the second pass will be for tags (and only when the record gets a new tag). At least, we are not sending the whole Resource (and its content) to be re-indexed two times :p

T-Dnzt:
@ptrikutam Not yet, since everything related to ES is a bit tricky, I want to add some feature specs first.

app/models/concerns/taggable.rb
T-Dnzt: After running a few tests, I still think we should keep the `cached_tags`, it will save us a tremendous amount of queries to the database and makes stuff faster.

T-Dnzt:
@ptrikutam This PR is now ready for a full review.

PR #130, opened 2017-02-16
Resource Upload + CRUD

T-Dnzt:
# Reason for Change

This pull request adds a way for users to upload resources (only URL type), listing the uploaded resources and managing them (basic CRUD).

# Changes

- Upload Resource: Logged in users can now click on `Upload Resource` in the dropdown menu to access a simple form (title + URL) to create a resource. That resource will be created with the current user as owner, the `resource_type` set to `url` and the privacy set to `priv` (private).
- My Resources: Just like `My Groups`, a user can access a page listing all the resources he has created. He can then view, edit or destroy them. 
- Privacy: A new enum was added to resources (`privacy`). It can either be set to `publ` or `priv` (default to `priv`) and can be used to prevent a resource from being seen by other users. Private resources are not indexed in ElasticSearch. `public` and `private` being reserved keywords in Ruby, I had to use `publ` and `priv` instead.
- Summary Card Changes: I started to improve the summary card design to make them more responsive. I'm planning to continue that effort in a future PR and apply it to all the summary cards. Some adjustments will also be made to some pages to improve the mobile experience.

# Notes

[Related Task](https://trello.com/c/quEqxdJF/13-green-commons-add-a-url-resource-upload-form)

### DIFF
--- a/app/views/resources/show.html.slim
+++ b/app/views/resources/show.html.slim
@@ -1,25 +1,23 @@
 .page-details
   .row.page-details__row
-    .col-xs-4.col-md-4
+    .col-xs-4.col-sm-4
       h6
         span class="glyphicon glyphicon--right glyphicon-#{resource_icon(@resource)}"
         = @resource.resource_type.upcase
-    
-    .col-xs-8.col-md-8
+    .col-xs-12.col-sm-8
       .resource-actions
         = render 'shared/components/btn_count', path: '#', 
                                                 text: 'ADD TO LIST', 
                                                 label: @resource.lists.count,
-                                                classes: 'resource-actions__button--right',
+                                                classes: 'resource-actions__button resource-actions__button--bottom resource-actions__button--right',
                                                 style: 'primary'
         - if policy(@resource).edit?
           = link_to 'EDIT', edit_resource_path(@resource), class: 'btn btn-primary resource-actions__button--right'
         - if policy(@resource).destroy?
           = link_to 'DELETE', @resource, method: 'DELETE', class: 'btn btn-default resource-actions__button--right', data: { confirm: 'Are you sure?' }   
-        
   .row.page-details__row.page-details__row--half
     .col-xs-12
-      h1
+      h1.page-details__title
         - if @resource.priv?
           span class="glyphicon glyphicon--right glyphicon-eye-close"
         = @resource.title
@@ -70,12 +68,12 @@
         span.glyphicon.glyphicon-menu-up.glyphicon--left aria-hidden='true'
                     
   .row.page-details__row
-    .col-xs-12.col-sm-2
+    .col-xs-12.col-sm-4
       h3 CONTENT
-    .col-xs-12.col-sm-10
-      .resource-actions.resource-actions__button--align-title
-        = link_to 'VIEW ORIGINAL', '', class: 'resource-actions__button btn btn-primary resource-actions__button--right'
-        = link_to '', class: 'resource-actions__button btn btn-primary' do
+    .col-xs-12.col-sm-8
+      .resource-actions
+        = link_to 'VIEW ORIGINAL', '', class: 'resource-actions__button btn btn-primary resource-actions__button--bottom resource-actions__button--right'
+        = link_to '', class: 'resource-actions__button btn btn-primary resource-actions__button--bottom' do
           | EXPORT
           span.glyphicon.glyphicon-cloud-download.glyphicon--left aria-hidden='true'
           
--- a/app/views/resources/show.html.slim
+++ b/app/views/resources/show.html.slim
@@ -3,7 +3,7 @@
     .col-xs-4.col-md-4
       h6
         span class="glyphicon glyphicon--right glyphicon-#{resource_icon(@resource)}"
-        = @resource.resource_type.capitalize
+        = @resource.resource_type.upcase
     
     .col-xs-8.col-md-8
       .resource-actions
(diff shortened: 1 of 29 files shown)


app/views/shared/_nav.html.slim
```diff
@@ -15,13 +15,15 @@ nav.navbar.navbar-inverse
               span.glyphicon.glyphicon-plus aria-hidden='true'
               span.caret
             ul.dropdown-menu role='menu'
+              li= link_to('Upload Resource', new_resource_path)
```
ptrikutam: Can we change this to say 'Add Resource'? I feel like "Upload" will indicate that we're uploading a file, when in reality we're just linking to a URL.

app/views/resources/show.html.slim
```diff
@@ -1,22 +1,26 @@
...
+        - if policy(@resource).edit?
+          = link_to 'EDIT', edit_resource_path(@resource), class: 'btn btn-primary resource-actions__button--right'
+        - if policy(@resource).destroy?
+          = link_to 'DELETE', @resource, method: 'DELETE', class: 'btn btn-default resource-actions__button--right', data: { confirm: 'Are you sure?' }   
   .row.page-details__row.page-details__row--half
     .col-xs-12
-      h1= @resource.title
+      h1.page-details__title
+        - if @resource.priv?
+          span class="glyphicon glyphicon--right glyphicon-eye-close"
```
ptrikutam: I'm not sure if this is the best place to keep this, I feel like this should probably be a bit more subtle. Can you add this icon to this location with the text "PRIVATE" next to it? https://cl.ly/1b0i2u2d3e3G

spec/models/resource_spec.rb
```diff
@@ -18,4 +18,39 @@
   end
 
   it_behaves_like 'indexable', :resource
+
+  describe 'privacy' do
+    let(:resource) { create(:resource, privacy: :priv) }
+
+    describe 'after_commit on: [:create]' do
+      it "adds the #{name} to the search index after creation" do
```
ptrikutam: I think the naming here is a bit misleading. Can you update this spec name to be something like `"does not add the #{name} to the search index after creation"`? Or am I mistaken about what the spec is intended to do?

spec/models/resource_spec.rb
```diff
@@ -18,4 +18,39 @@
...
+        allow(AddToIndexJob).to receive(:perform_async)
```
ptrikutam: Is this line needed if there is the line below?

spec/models/resource_spec.rb
```diff
@@ -18,4 +18,39 @@
...
+
+        expect(AddToIndexJob).not_to have_received(:perform_async).
+          with(resource.class.name, resource.id)
+      end
+    end
+
+    describe 'after_commit on: [:update]' do
+      it "updates the #{name} index in the search index after update" do
```
ptrikutam: Same as above. Can you update this spec name to be `"does not update the #{name} to the search index after update"`?

spec/models/resource_spec.rb
```diff
@@ -18,4 +18,39 @@
...
+
+        resource.touch
+
+        expect(UpdateIndexJob).not_to have_received(:perform_async).
+          with(resource.class.name, resource.id, [])
+      end
+    end
+
+    describe 'after_commit on: [:destroy]' do
+      it "removes the #{name} from the search index after deletion" do
```
ptrikutam: Same as above. Can you update the spec name?

spec/policies/resource_policy_spec.rb
```diff
@@ -9,11 +9,36 @@
...
   end
 
   context 'when user' do
     let(:user) { FactoryGirl.create(:user) }
 
     it { is_expected.to permit_action(:show) }
+    it { is_expected.to permit_action(:index)   }
+    it { is_expected.to permit_action(:new)     }
+    it { is_expected.to permit_action(:create)  }
+    it { is_expected.to forbid_action(:update)  }
```
ptrikutam: I'm a bit confused here -- what's the difference between the `owner` permissions and the `user` permissions?

In particular, if a user is not the owner of a resource, they shouldn't be able to update / edit / destroy, right?

ptrikutam (changes requested):
This is working & looking great, thanks for putting it together. Requested some minor changes and asked a q or two. Once yo umake those changes I'll approve. Thanks!

app/views/resources/show.html.slim
T-Dnzt: Here's the new version with your recommendations:

private

I think it looks alright like this, I wanted to use a lock instead of the eye but the one provided in Boostrap glyphicon looks terrible. Could add Font Awesome if you think the lock is better than the eye ;)

spec/models/resource_spec.rb
T-Dnzt: Yes, sorry about that. Thanks for spotting it!

spec/models/resource_spec.rb
T-Dnzt: Yup, else I'm getting:

```
#<AddToIndexJob (class)> expected to have received perform_async, but that object is not a spy or method has not been stubbed.
```

We need to setup the stubbing before the expectation.

spec/policies/resource_policy_spec.rb
T-Dnzt: The specs for `user` are:

```
    it { is_expected.to permit_action(:show) }
    it { is_expected.to permit_action(:index)   }
    it { is_expected.to permit_action(:new)     }
    it { is_expected.to permit_action(:create)  }

    it { is_expected.to forbid_action(:update)  }
    it { is_expected.to forbid_action(:edit)    }
    it { is_expected.to forbid_action(:destroy) }
```

Or can `show`, `index`, `new` and `create` and cannot `update`, `edit` and `destroy` with (`forbid_action`).

For `owner, it's:

```
it { is_expected.to permit_action(:show) }
it { is_expected.to permit_action(:index)   }
it { is_expected.to permit_action(:new)     }
it { is_expected.to permit_action(:create)  }
it { is_expected.to permit_action(:update)  }
it { is_expected.to permit_action(:edit)    }
it { is_expected.to permit_action(:destroy) }
```

Basically, an owner can do everything a user can do plus `update`, `edit` and `destroy`.

T-Dnzt:
@ptrikutam Ready for second review.

app/views/resources/show.html.slim
ptrikutam: I think the lock would be a great idea. Feel free to add font awesome!

spec/models/resource_spec.rb
ptrikutam: Strange -- even though the expectation is to *not* receive the stub. Oh well, works for me.

spec/policies/resource_policy_spec.rb
ptrikutam: D'oh. I completely misread the text. I didn't read the `forbid_` part of the `forbid_action`. Please disregard my comment :)

ptrikutam (approved):

ptrikutam:
LGTM 👍  great work!

Issue #123, opened 2017-02-03
Common Date Field for Groups/Resources/Lists/Users


T-Dnzt:
As mentioned in PR #117, we need to have a unique column containing the date that will be used to sort the records when different models are returned.

For all of them, we can use the date present in the metadata if there's one, or fall back to the `created_at` date. I guess we could call the field `date` or `published_at` maybe?

Let me know if you want me to proceed with this.

ptrikutam:
Let's go with `published_at`. Please proceed!

T-Dnzt:
I guess we can close this issue :)

PR #134, opened 2017-02-17
Generate published_at date after creation

T-Dnzt:
Related Issue: #123 

This pull request creates a shared date field named `published_at` for the `Group`, `Resource` and `List` models. This field is used to sort the records when searching through ElasticSearch. The value is defined when the record is created using either `metadata[date]` or `created_at`.

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -10,6 +10,8 @@ module Indexable
       yield
     end
 
+    after_create :set_published_at
+
     after_commit on: [:create] do
       run_if_public { AddToIndexJob.perform_async(self.class.name, id) }
     end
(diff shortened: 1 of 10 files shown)


app/models/concerns/indexable.rb
```diff
@@ -18,5 +20,19 @@ module Indexable
     after_commit on: [:destroy] do
       RemoveFromIndexJob.perform_async(self.class.name, id)
     end
+
+    def set_published_at
+      unless published_at
+        if defined?(metadata) && metadata && metadata['date']
+          begin
+            update_column(:published_at, metadata['date'])
+          rescue ActiveRecord::StatementInvalid
```
T-Dnzt: I had this method fail when the date stored in the metadata is not valid (like just a year for example), in that case I think we should default to the `created_at` date.

T-Dnzt:
@ptrikutam Ready for second review.

app/models/concerns/indexable.rb
ptrikutam: Hmm. We may need to address that in a future PR. I think we should be able to handle a year date for books, as many of them will only have a year, I believe. Can you create a separate issue and add it to GH?

T-Dnzt: @ptrikutam Issue Created: #137

PR #135, opened 2017-02-23
Add time stamping to user avatars

T-Dnzt:
# Reason For Change

There is a bug right now where it looks like a user cannot update his profile picture. This is due to the way the avatar filename is saved - since there is no stamping, the browser is caching the picture and won't get the new one because they have the same name.

# Changes

- Change the `filename` method using [this](https://github.com/carrierwaveuploader/carrierwave/wiki/How-to:-Use-a-timestamp-in-file-names) as an example.

### DIFF
--- a/app/uploaders/avatar_uploader.rb
+++ b/app/uploaders/avatar_uploader.rb
@@ -29,6 +29,11 @@ class AvatarUploader < CarrierWave::Uploader::Base
   end
 
   def filename
-    "#{model.full_name.parameterize}-avatar" if original_filename
+    @name ||= "#{model.full_name.parameterize}-avatar-#{timestamp}" if original_filename
+  end
+
+  def timestamp
+    var = :"@#{mounted_as}_timestamp"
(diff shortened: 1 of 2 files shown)


app/uploaders/avatar_uploader.rb
```diff
@@ -29,6 +29,11 @@ def extension_whitelist
   end
 
   def filename
-    "#{model.full_name.parameterize}-avatar" if original_filename
+    @name ||= "#{model.full_name.parameterize}-avatar-#{timestamp}" if original_filename
+  end
+
+  def timestamp
+    var = :"@#{mounted_as}_timestamp"
+    model.instance_variable_get(var) or model.instance_variable_set(var, Time.now.to_i)
```
T-Dnzt: Just reusing the sample code provided by Carrierwave here, in case you're wondering :D

ptrikutam: Thanks for clarifying1

PR #142, opened 2017-03-01
Add Disqus To Resources and Groups pages

T-Dnzt:
[Trello Task](https://trello.com/c/WQLZUjKt/20-green-commons-add-disqus-to-resource-group-pages)

This pull request adds the Disqus comments system to every resource and group pages. It also includes the comments count in the summary cards for these types of records.

### DIFF
--- a/app/views/shared/summary_cards/_group.html.slim
+++ b/app/views/shared/summary_cards/_group.html.slim
@@ -21,6 +21,11 @@
           = humanize_date(resource.published_at)
       .summary-card__metadata
         p
+          span.glyphicon.glyphicon-comment.glyphicon--right
+          span.disqus-comment-count data-disqus-identifier="#{url_for(resource)}" 
+            | 0 Comments
+      .summary-card__metadata
+        p
           span.glyphicon.glyphicon-user.glyphicon--right
           | #{resource.users.count} members
       .summary-card__metadata
@@ -28,6 +33,6 @@
           span.glyphicon.glyphicon-list.glyphicon--right
           | #{resource.lists.count} Lists
       .summary-card__metadata.summary-card__metadata--right
-        = link_to 'Read More >>', resource
+        = link_to 'Read More', resource, class: 'btn btn-default'
       .clearfix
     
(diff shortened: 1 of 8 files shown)


app/views/shared/summary_cards/_group.html.slim
```diff
@@ -21,13 +21,18 @@
           = humanize_date(resource.published_at)
       .summary-card__metadata
         p
+          span.glyphicon.glyphicon-comment.glyphicon--right
+          span.disqus-comment-count data-disqus-identifier="#{url_for(resource)}" 
+            | 0 Comments
```
ptrikutam: > It also includes the comments count in the summary cards for these types of records.

Is this the line that's supposed to do it? I've added a comment on a resource locally but I'm not seeing the count increase...

T-Dnzt:
Whoops, this is because resources are private by default. When I added the column with default value, they were all set to `priv`. You need to run the following command in the `rails console`:

`Resource.where('user_id IS NULL').update_all(privacy: 'publ')`

I can also make it a rake task if you prefer. I created [this issue](https://github.com/greencommons/commons/issues/143) so we don't forget to update the ETL to  set the privacy for the next imports.

app/views/shared/summary_cards/_group.html.slim
T-Dnzt: Yeah that's the line... I'm having some issues too. It works correctly for the first comment but then stops incrementing. Anyway, I'm going to debug it but that's the kind of reasons why I dislike using pre-made/external tools :D

T-Dnzt: @ptrikutam Just spent some time trying to figure it out, and I can't find why it's not working (and it's starting to piss me off :)). I read countless posts and tried many solutions, but it either doesn't work or just show `1 comment` (even when the resource has more). 

My last hope is that it might be related to the fact that I'm using the staging URL in the configuration (can't put localhost). We might have to try in staging and see how it goes (really don't think it's going to help, but I'm running out of ideas...).

ptrikutam: OK, for the time being, don't worry about this. This isn't a dealbreaker and we shouldn't have it block the PR.

Issue #139, opened 2017-03-01
Search API


ptrikutam:
### Overview

We'd like to start building out an external-facing JSON RESTful API. Let's start the discussion & development of said API for the Search endpoint. We're choosing this endpoint because it's probably the most useful, and also wont require any kind of authentication mechanism.

#### General API Requirements

- For the time being, don't worry about authenticating or verifying the requester. Anyone can hit this endpoint and get results.
- The API should return a JSON response.
- The API should be versioned (putting everything under v1 is good enough for now).

#### Search API Requirements

These requirements are presented in order of priority:

- First, allow searching via a query string. At its simplest, the API should return a JSON response for a request like this: `GET https://greencommons.herokuapp.com/api/v1/search.json?q=<query>`
- The API should allow for paginating the results
- The API should allow for sorting the results
- We should allow faceting via the API (i.e. filter by Resource, Group, Article, etc)

### Open to discussion

- I'm open to how we implement it (i.e. using a gem like [Grape](https://github.com/ruby-grape/grape), or manually)
- I think the API should be logically separated somehow from the controllers, but I haven't fully thought through what that looks like. Got any suggestions on how best to go about this?
- How should we allow for additional parameters in the search API for faceting, pagination, etc? Should we accept a JSON payload in a POST or should we accept query parameters?

T-Dnzt:
@ptrikutam For the JSON format, I was thinking about going with a defined specification: [JSON API](http://jsonapi.org/). That would save us some time for the documentation later on, and it's always better to use some specification/standard than making our own format, especially for a public facing API.

What do you think?

T-Dnzt:
- I'm open to how we implement it (i.e. using a gem like Grape, or manually). I think the API should be logically separated somehow from the controllers, but I haven't fully thought through what that looks like. Got any suggestions on how best to go about this?

I think it's simpler to just use Rails controller namespaced into something like `Api::VX`.

- How should we allow for additional parameters in the search API for faceting, pagination, etc? Should we accept a JSON payload in a POST or should we accept query parameters?

This should be done using query parameters in my opinion.

`GET https://greencommons.herokuapp.com/api/v1/search.json?q=<query>`

[...]

PR #146, opened 2017-03-23
Web API Basic Setup

T-Dnzt:
This pull request adds the basic setup of the Green Commons Web API. It follows the JSON API to render the responses.

- [Trello Card](https://trello.com/c/NmWE5A0t/22-green-commons-search-api-basic-setup)
- [GitHub issue](https://github.com/greencommons/commons/issues/139)

# Changes

## JSON API 

In order to support JSON API, I first tried using some of the existing gems like Active Model Serializers. Sadly, I wasn't able to make them work with the requirement for the search: being able to return different types of model in the same response.

After fighting with these gems, I decided to switch to a library I built a while back that serialize data following the JSON API spec. I had to make a few changes to it to accommodate the needs.

## Yumi

That library is called Yumi (meaning "archery bow", because we're shooting JSON at people) and is composed of a set of presenters and helpers to make it easy to create presenters for our models.

I don't think the code in the library itself requires review, but feel free to check it if you have time. I'm planning to extract the code into a gem soon - after that I will just include the gem in the project. For now, I'm keeping it in `lib/` in order to be able to make changes easily while I'm building the API and finding things to improve.

Yumi gives us the option to create presenters like [this one](https://github.com/greencommons/commons/blob/173c5f6eefc8902e527f8fd476372f8f15707d9b/app/presenters/v1/group_presenter.rb) for our models and use this code to render them:

```
group = Group.find(params[:id])

Yumi::Presenter.new(url: base_url,
                    current_url: request.original_url,
                    resource: group,
                    includes: [],
                    presenters_module: ::V1).as_json_api
```

## The Search Controller

The regular search controller logic has been extracted into a new class to be reused by the search API: [SearchBuilders::Search](https://github.com/greencommons/commons/blob/173c5f6eefc8902e527f8fd476372f8f15707d9b/app/services/search_builders/search.rb)

## Other API controllers

I created a simple controller for groups/resources/lists with only the `show` action for now. More actions will be added later, but since we have links inside the JSON representations, I wanted some of them to work as expected (still have to fix the relationship links).

## Pagination, filtering, etc

These should be working but can be improved. I will do that in a future PR which is why there is no tests for them yet. I need to see how to have the pagination for records found from ElasticSearch and from the database work in the same way for example.

## You may also like

This will be added as a new key to the JSON for the search endpoint (`related`) and will be added in a future PR as well.

## Documentation

How do you want me to write the public documentation for the API? Any specific format/tool in mind?


spec/requests/api/v1/search_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::GroupsController, type: :request do
+  describe 'GET /api/v1/groups/:id' do
```
ptrikutam: May want to update this spec name.

spec/requests/api/v1/search_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      create(:list, name: "#{title} My List")
+
+      wait_for do
+        Elasticsearch::Model.search(title, [Resource, Group, List]).results.total
+      end.to eq(3)
+
+      get "/api/v1/search?q=#{title}"
+    end
+
+    it 'returns the group as JSON API', :worker, :elasticsearch do
```
ptrikutam: Same here -- might want to update the spec name.

app/models/concerns/indexable.rb
```diff
@@ -4,6 +4,7 @@ module Indexable
   included do
     include Elasticsearch::Model
     index_name SearchIndex.index_name(self)
+    attr_accessor :relevancy
```
T-Dnzt: Yes exactly, it's just an attribute that will be set after we get the ES results and before we pass the objects to the presenters.

PR #148, opened 2017-04-05
Web API - Sorting 

T-Dnzt:
[Trello Task](https://trello.com/c/nlWlTplO/24-green-commons-search-api-sorting)

Depending on [the pagination PR](https://github.com/greencommons/commons/pull/147). Switch base branch to master once that PR has been merged.

# Changes

This pull request implements sorting for the web API Search resource using `published_at`, `created_at` and `updated_at`. I'd like to add sorting by name but it would require changing the name of the field of the `Resource` model from `title` to `name`. In theory, it's possible without that change, but I wasn't able to make it work.

This PR also include the switch to sorting in the JSON API style using `published_at` (ascending) or `-published_at` (descending).


T-Dnzt:
@ptrikutam Can you try to re-index all the records? It's probably due to the fact that the `published_at` field was added after your previous indexing.

ptrikutam:
@T-Dnzt Looks like it's working! The issue was that, even though I was re-indexing the records, all the `published_at` dates were still null (since that was added after the records were imported into my local database).

I just ran 

```
Resource.all.each do |resource|
  resource.set_published_at
end
```

Then, re-indexed everything and it worked fine.

PR #154, opened 2017-04-22
New Design for Search Page

T-Dnzt:
# Reason for change

# Changes

This pull request introduces the first changes of the new design: the search page (as well as the header, footer and set of styles to reuse in future pages). 

- New header
- New footer
- New search page
- New filtering system that auto-reloads when a checkbox is ticked or the sorting select is changed
- Setup the tests to work with JS (in order to test the filters)
- Fix all the tests to use JS

# Note

Not all the filters have been implemented. I added the design to filter by tag and date but these are not working yet. Also, the `clear` links have not been implemented. This will be done in a different pull request to avoid having this one get too big. I only fixed the other filters in order to have all the current tests passing.

The design of most of the other pages is completely broken right now, so we probably shouldn't merge this until the next pull requests can be merged as well.

If you're having problem indexing the documents, you can run the following rake task (temporary fix until we fix the `content` field:

```
rake resources:clean_resources
```

It will remove empty date (from metadata) and set the content to string if it's not one already.

# Responsive Design

While the website works properly for small screens, it doesn't look great. I don't want to implement something myself so it would be good to have some designs for phone/tablet sizes :).

# Screenshot

Here's how it looks like:

screen shot 2560-04-24 at 6 35 33 pm


spec/support/wait_for_ajax_helpers.rb
```diff
@@ -0,0 +1,32 @@
+# This module waits for the Capybara default wait for AJAX requests. Once that is
```
ptrikutam: This looks familiar :)

Glad it's of use here.

app/views/search/_filters.html.slim
```diff
@@ -0,0 +1,64 @@
...
+            li.filters__item
+              = render 'search/filter', name: :reports, category: :resource_types
+        li.filters__item
+          = render 'search/filter', name: :lists, category: :model_types
+        li.filters__item
+          = render 'search/filter', name: :groups, category: :model_types
+      .clearfix
+      
+  .filters
+    .filters__title.text-xs-left.text-center
```
T-Dnzt: Removing it completely for now, else the query params will keep showing up. Super easy to add back anyway.

app/views/search/_filters.html.slim
```diff
@@ -0,0 +1,64 @@
...
+        li.filters__item
+          = render 'search/filter', name: :Downloadable, category: :model_types
+      .clearfix
+      
+  .filters
+    .filters__title.text-xs-left.text-center
+      h6
+        | Tags
+        span.filters__clear= link_to 'Clear', root_path, class: 'gray-light pull-right'
+    .tags-list
```
T-Dnzt: I'm already getting the list of all the tags present in the search results (to provide the suggestions). We will need to change/add a bit of logic to get the top tags and count the occurrences. But I'm wondering what happens when a user clicks on a tag?

app/views/search/_results.html.slim
```diff
@@ -0,0 +1,27 @@
+.col-xs-12.col-sm-9
+  .page-details__row--quarter
+    = form_tag search_path, method: :get do
+      .row
+        .col-xs-12.col-sm-6
+          h6
+            span= "Results for \"#{@query}\" "
+            span.gray-light
+              | &bull;
+              = " #{number_with_delimiter(@total_count, delimiter: ' ')}"
```
T-Dnzt: I can change it but the current design reflects what I found in the Sketch files:

screen shot 2560-04-27 at 12 14 44 pm

app/models/concerns/indexable.rb
```diff
@@ -33,13 +33,15 @@ def set_published_at
       unless published_at
         if defined?(metadata) && metadata && metadata['date']
           begin
-            update_column(:published_at, metadata['date'])
-          rescue ActiveRecord::StatementInvalid
-            update_column(:published_at, created_at)
+            self.published_at = Date.parse(metadata['date'])
+          rescue StandardError
```
T-Dnzt: I changed it to 'StandardError' in order to rescue from both the `ActiveRecord::StatementInvalid` and the `Argument Error: Invalid Date` but then I extracted the `save!` out of the rescue block (the `ActiveRecord::StatementInvalid` shouldn't happen anymore anyway). I forgot to switch back and I'm going to do that right now ;)

app/models/concerns/indexable.rb
```diff
@@ -33,13 +33,15 @@ def set_published_at
...
+            self.published_at = created_at
```
T-Dnzt: Hmmm, the only way I can think of right now is to set the date to the January 1st if only the year is available. Do you think that would be good enough for the search results? Probably better than using the `created_at` anyway.

app/helpers/score_helper.rb
```diff
@@ -1,11 +1,12 @@
...
-    if score > 1
-      'label-success'
-    elsif score > 0.5
-      'label-warning'
+  def score_opacity(score)
+    case score
+    when 0..1
+      score / 2
+    when 1..10
+      score / 10 + 0.5
```
T-Dnzt: I just thought it would be too hard to see the number for very low scores. I don't mind removing it - also if you have a better idea on how to handle this, I'm all ears :D Since I can't really know the limits of the score, it's quite hard to find a good way to show it.

app/services/suggesters/tags.rb
```diff
@@ -9,12 +9,17 @@ def initialize(tags:, except: [], limit: 10, models: [Group, List, Resource])
...
     def suggest
       load_es_params
-      except_records if @except
-      Elasticsearch::Model.search(@es_params, @models).records
+      records
     end
 
     private
 
+    def records
```
T-Dnzt: This is used to load the suggestions on the search page or on the resource page. It will loads records based on the passed tags (and the other options). The `except` is only used for the resource page, in order to remove the current resource from the list of suggestions.

T-Dnzt:
@ptrikutam Thanks for all the feedback! I think I fixed everything except the results count with delimiter. Can you confirm that you want me to change it even though it will differ from the Sketch design?

Let me know if I forgot anything ;)

app/views/search/_filters.html.slim
ptrikutam: Good question. I'd say it should either filter the current list or trigger a new search. Not sure exactly which one just yet.

app/views/shared/summary_cards/_resource.html.slim
```diff
@@ -1,66 +1,42 @@
...
-          .summary-card__more
-            = link_to resource, method: 'DELETE', class: 'btn btn-default', data: { confirm: 'Are you sure?' } do
-              span class="glyphicon  glyphicon-remove"
+    .summary-card__row--double
+      div class="entity-icon entity-icon--#{resource.resource_type}"
+        i class="fa fa-#{resource_icon(resource)}"
+      .summary-card__title
+        h5= link_to resource.title, resource
+        h6
+          - if resource.user&.full_name&.present?
```
ptrikutam: @T-Dnzt Can you make this the author, not the user? It should be in `resource.metadata['creators']` I believe.

app/views/shared/summary_cards/_group.html.slim
```diff
@@ -1,38 +1,32 @@
...
-          span.glyphicon.glyphicon-calendar.glyphicon--right
-          = humanize_date(resource.published_at)
-      .summary-card__metadata
-        p
-          span.glyphicon.glyphicon-comment.glyphicon--right
+        = resource.short_description
+    = render 'shared/components/tags_list', tags: resource.cached_tags
+    .summary-card__row.summary-card__row--last
+      .summary-card__more.summary-card__more--space.summary-card__more--buttons
+        = link_to 'Add to a list (25)', root_path, class: 'btn btn-gc'
```
ptrikutam: Not sure where the CSS resides but can you remove the shadow on this button: https://cl.ly/3F0J1n3n1Z3D

ptrikutam: Please remove it for all the other summary card types too.

app/views/shared/_nav.html.slim
```diff
@@ -8,26 +8,38 @@ nav.navbar.navbar-inverse
...
+        li
+          = form_tag search_path, method: :get, class: 'navbar-form customer-search-form' do |form|
+            .form-group.has-feedback
+              = text_field_tag :query, params[:query], placeholder: 'Search...', class: 'form-control navbar__search-input'
+              = button_tag type: 'submit', name: nil, class: 'form-control-feedback navbar__search-button' do
+                i.glyphicon.glyphicon-search
+
+        - if user_signed_in?
+          li
+            .btn-group.navbar-btn
```
ptrikutam: Can you remove the shadow from this button? https://cl.ly/3A2e3u0X3524

app/views/search/_results.html.slim
```diff
@@ -0,0 +1,27 @@
...
+        .col-xs-12.col-sm-2.col-sm-offset-4
+          = hidden_field_tag :query, @query
+          = select_tag :sort, 
+                       options_for_select([ ['Score', 'score'], 
+                                            ['Recent First', '-published_at'],
```
ptrikutam: Small request: can you change it to "Newest first"

app/views/search/_results.html.slim
ptrikutam: For sure, that was a mistake. Can you add commas?

app/assets/javascripts/filters.js
```diff
@@ -0,0 +1,70 @@
...
+      dataType: 'script'
+    });
+  };
+
+  $('input[name="daterange"]').daterangepicker({
+    "startDate": moment().subtract(7, 'days'),
+    "endDate": moment(),
+    "buttonClasses": "btn btn-sm btn-gc",
+    locale: {
+      format: 'DD/MM/YYYY'
```
ptrikutam: Can you change the format to the American format? `MM/DD/YYYY`. Only because the initial audience will be from the US :)

app/assets/stylesheets/components/_nav.scss
```diff
@@ -0,0 +1,46 @@
+.navbar {
```
ptrikutam: Can you remove the box-shadow here: https://cl.ly/1x3M0J3I1909

app/assets/stylesheets/application.scss
```diff
@@ -1,14 +1,27 @@
...
 @import "helpers/fonts";
 @import "helpers/text-aligning";
 @import "helpers/images";
+@import "helpers/buttons";
+@import "helpers/forms";
+@import "helpers/size";
 
+@import "components/nav";
+@import "components/pagination";
+@import "components/daterangepicker";
```
ptrikutam: Wasn't sure where to put this, but a few request here:

1. Can you remove the shadows from the buttons: https://cl.ly/3M3B3U0C2V1h
2. Can you change the background of the daterange picker to be white: https://cl.ly/26403z1U303b
3. Can you reduce the font size to match: https://cl.ly/26403z1U303b
4. Add some spacing to the top of the dropdown so that it's a few pixels below the input: https://cl.ly/291S1C0U3S16. It should look like: https://cl.ly/2f1B3W0j0X40

ptrikutam: 5. Please decrease the border-radius of the daterange picker input to 6px.

app/views/search/_results.html.slim
T-Dnzt: Should I also add `results` after? I'm not sure because we already have `Results for` right before sooo, let me know ;)

app/views/shared/summary_cards/_resource.html.slim
T-Dnzt: Yes sure, sorry about that ;)

app/assets/stylesheets/components/_summary-card.scss
```diff
@@ -3,7 +3,7 @@
   width: 100%;
 
   &__body {
-    padding: 15px;
+    padding: 25px;
     width: 100%;
     @include box-shadow(0 1px 4px rgba(0,0,0,.3));
```
T-Dnzt: Note that I used `#dfe5e8` for the border color because it's a variable already defined for other borders/lines (got it from the design). In the Sketch file, the border color is actually `#5f7d8c` at 0.2 opacity but I thought it was better not adding another variable for that because the two colors are pretty close (also keeping it consistent with other borders).

T-Dnzt:
@ptrikutam I made the requested changes ;) Let me know if there's anything else!

app/assets/stylesheets/components/_summary-card.scss
ptrikutam: Sounds good.

ptrikutam:
Looks amazing. Great work 👍

PR #155, opened 2017-05-01
New Design for Resource Page

T-Dnzt:
# Reason for change

# Changes

This pull request includes the new design for the resource page, with the colors changing depending on the type of resource being viewed (report, article, book, url). I had to make a few choices for the edit/delete buttons that were not in the design as well as the form to add tags. I also included the creators (only the owner of the resource was being shown in the design) and the privacy of the resource in the metadata list. 

The discuss section was not implemented and left with the Disqus design.

# Note

The color used for URL resources is the same one as the one for books since there was no color for it in the design.

# Screenshot

Here's an example of how it looks like with all the buttons showing:

screen shot 2560-05-01 at 10 14 59 pm


app/assets/stylesheets/theme/_variables.scss
```diff
@@ -34,6 +34,29 @@ $brand-info:            #9C27B0 !default;
 $brand-warning:         #ff9800 !default;
 $brand-danger:          #e51c23 !default;
 
+$books-clear:           #f3faee !default;
```
ptrikutam: Very small nitpick: rather than naming it `-clear`, perhaps `-light` might be a good name? Clear implies that it is transparent.

app/assets/javascripts/components/tags.jsx
```diff
@@ -40,26 +40,21 @@ var Tags = React.createClass({
...
-                     className="form-control page-details__tag-input" />
-            </div>
-            <div className="col-xs-4 col-md-4">
-              <button id="add_tag_button" name="button" type="submit" className="form-control btn btn-primary">
-                <span aria-hidden="true" className="glyphicon glyphicon-plus"></span>
-              </button>
-            </div>
-          </form>
-        </div>
+      <div className="page-details__tag-form">
```
T-Dnzt: Indeed, this was a bug - fixing it now ;)

PR #158, opened 2017-05-06
New design for lists

T-Dnzt:
# Reason for change

# Changes

- Implement lists with the new design
- Add tests for lists
- Add owner select box
- Fix button positions in groups and resources
- Add tags input
- Change alert colors

# Add to list

Since this PR is getting a bit big, I'm going to keep the "Add to list" separated. Coming soon! I'm also planning to change the select box in the list form to pull the results when the user via AJAX.

### DIFF
--- a/app/views/shared/summary_cards/_list.html.slim
+++ b/app/views/shared/summary_cards/_list.html.slim
@@ -14,10 +14,8 @@
     .summary-card__row
       p
         = resource.description
-        | Lorem ipsum lorem ipsum lorem ipsum lorem ipsum 
-        | Lorem ipsum lorem ipsum lorem ipsum lorem ipsum
     .summary-card__row.summary-card__row--last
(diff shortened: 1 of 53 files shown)


app/views/shared/summary_cards/_list.html.slim
```diff
@@ -4,7 +4,7 @@
       div class="entity-icon entity-icon--lists"
```
ptrikutam: For some reason, tags aren't showing up in the summary card. I've got them listed here: https://cl.ly/0b3G1K3V272F

But they don't show up in the search results: https://cl.ly/2Y1i3K0G0G3v

T-Dnzt: @ptrikutam Indeed, I forgot to include the partial showing the tags in the list summary card! Thanks for catching that.

PR #159, opened 2017-05-11
"Add To List" Feature

T-Dnzt:
# Reason for change

# Changes

This pull request includes the following changes:
- Add to list feature
- Remove from list feature
- Items sorting in List
- Improving the autocomplete system and implementing list autocomplete

# Notes

- I'm not sure about who can add/remove something to/from a list, so for now I let every [logged in user](https://github.com/greencommons/commons/pull/159/files#diff-124f5c108af748388e59405611184889R11) but this should be changed. 
- I wasn't able to write a test for the "add to list" feature... Everything was going okay in the test until the submit and I haven't been able to figure out what's wrong exactly while everything is fine in the browser. I already spent a lot of time trying to figure this out but I wasn't able to find a solution yet. Want me to dig more?
- Before deploying this to production, the ElasticSearch addon on Heroku  needs to be upgraded to the 5.1+ version.


app/views/lists/_items.html.slim
```diff
@@ -0,0 +1,27 @@
...
+          
+      .row.page-details__row--double
+        .col-xs-12
+          - @items.each_slice(2) do |slice|
+            .row
+              - slice.each do |list_item|
+                .col-xs-12.col-md-6
+                  = render 'shared/summary_cards/main', resource: list_item.item, 
+                                                        minimalist: true, 
+                                                        remove: list_item_path(list_item, sort: @sort, page: @page)
```
ptrikutam: Can we only show this if we're the list owner?

T-Dnzt: I'm gonna use the same logic than for the rest: if the user owns the list or is an admin of the group which owns it - is that okay?

T-Dnzt: Or does any member of a group can do it?

T-Dnzt:
@ptrikutam 

> Currently, you can add to any list in the database. Ideally, a user should only be able to edit a list or add to a list if they're a direct owner, or if they're part of a group which is the owner of a list

This has been changed in my latest push. However, only group admins can edit the list the group owns. It's super easy to change if you want tho, but I think it makes more sense, maybe?

> For the autocomplete dropdown when adding to a list, it'd be great to display all the lists a user directly owns, as well as maybe the first 5 lists from groups a user belongs to: https://cl.ly/0S1B2i1r3d1c

Can I do that one in a separate PR? I need to get one of the other PR merged first, it has some stuff that I need to reuse to inject some options in the select.

> On the group page, we should see the lists associated with that group.

I'm seeing them, can you give me more info? Maybe I'm missing something (the buttons are not working tho and we don't have a page yet to see all the lists for a group):

screen shot 2560-05-22 at 5 27 28 pm

> Lists marked "private" should only be viewable by those who are either direct owners or members of the group it is owned by.

This should be fixed now. I also fixed a bug where a list would stay indexed in ES even if switched to private.

> A logged out user should not be able to add anything to a list (Add to list might just have to do nothing... Or, alternatively, if we want to get fancy, we should show a different message that says "You have to be logged in to do that").

I added the "You have to be logged in to do that" message, took just 5 minutes :p.

app/views/lists/_items.html.slim
ptrikutam: @T-Dnzt any member of a group can do it.

ptrikutam:
@T-Dnzt 

> Can I do that one in a separate PR? I need to get one of the other PR merged first, it has some stuff that I need to reuse to inject some options in the select. I created a new task in Trello.

Sure, no problem.

PR #162, opened 2017-06-02
Improve Add To List dropdown

T-Dnzt:
# Reason For Change

Currently the dropdown to add a resource to a list doesn't show any list. It would be nice to show 5 lists owned by the current user to make things easier.

# Changes

- Switch to Selectize from Select2 in order to correctly pre-populate the select
- Allow any user belonging to a group owning a list to edit it
- Change the add to list feature to only allow a user to add to lists he owns or belongs to a group owning them

### DIFF
--- a/app/controllers/autocomplete_controller.rb
+++ b/app/controllers/autocomplete_controller.rb
@@ -7,8 +7,11 @@ class AutocompleteController < ApplicationController
   end
 
   def lists
+    existing_lists = current_user.group_owned_lists.pluck(:id)
+    allowed_lists = current_user.owned_lists.where.not(id: existing_lists) +
+                      current_user.group_owned_lists.where.not(id: existing_lists)
     lists = Suggesters::Lists.new(query: params[:q],
-                                  except: current_resource.lists).suggest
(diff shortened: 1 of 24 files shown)


app/controllers/autocomplete_controller.rb
```diff
@@ -7,8 +7,11 @@ def members
   end
 
   def lists
+    existing_lists = current_user.group_owned_lists.pluck(:id)
```
ptrikutam: Not sure I follow what's happening here. Why not include these?

T-Dnzt: @ptrikutam Whoops, wrong one - this should be `current_resource.lists.pluck(:id)` because we don't want the lists to which the current resource already belongs to. Fixing it.

T-Dnzt:
@ptrikutam 

```
Caused by: java.lang.IllegalArgumentException: no mapping found for field [name_suggest]
```

This means your indexes don't have the latest version of the mapping and you need to recreate them and re-index the documents ;) 

I added suggestions when creating a list with the current user and the five most recent groups that user has joined.

PR #163, opened 2017-06-08
Web API - New Group Resources

T-Dnzt:
# Reason for change

# Changes

- Implementation of the authentication mechanism using a custom HTTP Scheme
- Add index, create and update for groups through the API
- Add API Keys available in a user's profile


T-Dnzt:
@ptrikutam Will get those fixed in a different PR :) Already created a task in Trello for it.

PR #164, opened 2017-06-08
Web API - Group Members Resources

T-Dnzt:
# Reason for change

# Changes

- Allow retrieval of group members
- Allow addition of new members (including the current user) to a group (if group admin)
- Allow removal of members (including the current user) from a group (if group admin)


T-Dnzt:
@ptrikutam Sorry, I made a mistake while writing the curl example. It should be:

```
curl http://localhost:3000/api/v1/groups/:id/relationships/users
```

PR #166, opened 2017-06-10
Web API - Resources endpoints

T-Dnzt:
# Reason for change

# Changes

This pull request adds new endpoints for resources:

- View a resource
- Create a resource
- Update a resource

# Note

These endpoints work exactly like in the web application, which means resources are created as `private` by default. 

- Should we make it possible for users to change all list attributes through the API and upload files?
- I haven't done anything with tags yet, I'm thinking about just doing it in a common PR for groups/resources/lists. Should a user just be able to send a list of tags and we add them?

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,44 +213,28 @@ curl -g -X GET 'https://greencommons.herokuapp.com/api/v1/search?q=wind&filters[
 
 ##### Retrieve a resource - `/api/v1/resources/:id`
 
-__Example__
-
 ```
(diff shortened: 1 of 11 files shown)


ptrikutam:
> Should we make it possible for users to change all list attributes through the API and upload files?

For the time being, no let's not worry about this.

> I haven't done anything with tags yet, I'm thinking about just doing it in a common PR for groups/resources/lists. Should a user just be able to send a list of tags and we add them?

Let's hold off on this one as well.

PR #168, opened 2017-06-10
Fix lists not showing up correctly on group page

T-Dnzt:
# Reason for change

Currently, the group page isn't displaying the lists owned by the group. Instead, it shows the lists to which a group has been added.

# Change

This PR changes the logic to display the lists owned by the group being viewed instead of the lists to which the group belongs.


T-Dnzt:
@ptrikutam Thanks, good catch!

PR #169, opened 2017-06-13
Prepare new resource ingestion

T-Dnzt:
# Reason For Change

New resources have been added in S3 and need to be pulled and inserted in the application DB.

# Changes

- Create a new ETL config to pull files from `validcommons`.
- Add new resource types.
- Add two new columns to resources (`short_content` and `long_content`). The current `content` column will be removed in a future PR.
- Create tags for resources.
- Add a rake task to transfer the resources `content` to `long_content`.


gtourtellot:
Yes, it makes sense to overwrite (or, say, first delete) everything that's
in the database now.

T-Dnzt:
@ptrikutam @gtourtellot Can I really just delete all the resources currently in the production database? We're keeping groups, lists and users though, right?

gtourtellot:
The previous group data is going away; what had been there previously is
now reworked as resources of 'profile' type.  We've renamed 'groups' as
'networks', though it's the still the same idea.  For networks, we now have
a schema, slightly modified from what was groups; see:
 'network_array_schema.json' in s3.  There is a small amount of network
data at this time, also on s3 now.

@ptrikutam <https://github.com/ptrikutam> can speak better as to how to
treat existing list data.

PR #178, opened 2017-07-20
Use short/long content appropriately

T-Dnzt:
# Reason for change

If no short content is available, we should show the truncated long content.

# Changes

- Use `short_content` if available, else truncate `long_content`

### DIFF
--- a/app/views/shared/summary_cards/_resource.html.slim
+++ b/app/views/shared/summary_cards/_resource.html.slim
@@ -27,13 +27,11 @@
       .clearfix
     .summary-card__row
       p
-        - if resource.url?
-          = link_to resource.url, resource.url, target: '_blank'
+        - if resource.short_content
+          = resource.short_content.truncate(400)
         - else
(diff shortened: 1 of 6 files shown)


app/views/shared/summary_cards/_resource.html.slim
```diff
@@ -23,13 +23,11 @@
       .clearfix
     .summary-card__row
       p
-        - if resource.url?
-          = link_to resource.url, resource.url, target: '_blank'
```
ptrikutam: Out of curiosity, why were the URLs being shown in this section instead of the description? Also, removing this line won't remove the "View Original" link, correct?

T-Dnzt: Because URLs resources can't have descriptions yet. When you create a new one, you can only give it a name and a URLs. I think we didn't have the View Original button at that time so I was showing the URLs as a description ;) The "View Original" link won't be removed.

PR #181, opened 2017-07-28
Fix API & API documentation

T-Dnzt:
[Trello Card](https://trello.com/c/jlFaPP9R/97-green-commons-separate-out-fix-api-documentation)

# Reason For Change

GC requested the API documentation to be updated and improved.

# Changes

- Move documentation to the `API.md` file
- Replace occurrences of `localhost` with the production URL
- Replace schema URLs (`/:id`) with database values (`/1`)
- Re-test all curl requests
- Fix relationship IDs
- Fix generated URL when query parameters are submitted


API.md
```diff
@@ -0,0 +1,259 @@
...
+     -X POST \
+     -H 'Authorization: GC SOcdhU1l5Jiebg6EDtyDFQ:1168c1433a2e7aa5945401e7245186e3' \
+     -H 'Content-Type: application/vnd.api+json' \
+     -d '{ "data": { "type": "users", "attributes": { "email": "john@example.com", "password": "password", "password_confirmation": "password"} } }'
+```
+
+##### Update user
+
+```
+curl http://greencommons.herokuapp.com/api/v1/users/10 \
```
ptrikutam: Perhaps it's not best to use User 10 in this, since that's admin@greencommons.org. Then again, that's probably not going to end up being the admin account in the long term.

@gtourtellot -- Thibault has raised a good point in that specifying actual IDs in the documents may not be the best idea, because if we ever update the database or those IDs change, the documentation becomes inaccurate. Perhaps rather than providing working examples, we leave these parts generic since we don't have a dynamic API documentation system set up. What do you think?

T-Dnzt: @ptrikutam Since I'm using the keys from that account to connect to the API, it's the only one I can update (unless I create a new account and update all the access/secret keys).

Waiting on the conclusion about using actual IDs or not before doing it ;)

gtourtellot: I was imagining that the numbers would just be arbitrary, not working examples.  There could be a sentence in the doc explaining that.  I think it'd be pretty easy for a user to grasp that; alternatively, seeing something like ":id" in place of a number seems confusing.

Let me know if you think there's a better approach.

ptrikutam: Gotcha. OK -- @T-Dnzt, you can keep the numbers / IDs as is for now. As for the actual token, can you put in dummy access/secret values so people don't make changes with that specific account?

You can also add in a line or two about replacing the authentication as well as the IDs in the examples with users own.

T-Dnzt:
@ptrikutam Ready for another look ;)

PR #184, opened 2017-08-10
Add About page

T-Dnzt:
This pull request adds the about page as requested by the GC team.

### DIFF
--- a/app/views/static_pages/about.html.slim
+++ b/app/views/static_pages/about.html.slim
@@ -1,0 +1,91 @@
+.page-details--creme-bg
+  .container
+    = render 'shared/alerts'
+    .row
+      .col-xs-12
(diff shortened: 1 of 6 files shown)


app/views/static_pages/about.html.slim
```diff
@@ -0,0 +1,91 @@
...
+              span.form-box__underline build
+              = ' '
+              | networks; and to
+              = ' '
+              span.form-box__underline work collaboratively
+              = ' '
+              | . We aim to provide users the information and tools that will help them create a
+              | more sustainable world.
+
+            h5 Vision
```
ptrikutam: Can we add a bit of margin above the h5? Also, can you make it a bit bigger? Perhaps an h4 or even h3?

PR #187, opened 2017-08-24
Implement date filtering

T-Dnzt:
# Reason for change

Date filtering is currently not implemented, this PR adds it.

# Changes

- Add date filtering
- Add tests to ensure dates can be selected and URLs containing the filters are correctly interpreted by the calendar

### DIFF
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -21,7 +21,7 @@ var initFilters = function() {
       end = moment.unix(daterange.data('end'));
     } else {
       start = moment().subtract(7, 'days');
-      end = moment()
+      end = moment();
     }
 
     daterange.daterangepicker({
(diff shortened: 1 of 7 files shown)


T-Dnzt:
@ptrikutam The highlighted random date is not actually a random date, it's just "today" being highlighted in the following month because that month is showing the last week of the previous month at the beginning. In the screenshot you shared, it's August 28 being shown both in the August month but also in the September month since the first week of September starts on September 1st. Does it make sense? I'm not sure how I can prevent this.

ptrikutam:
> The highlighted random date is not actually a random date, it's just "today" being highlighted in the following month because that month is showing the last week of the previous month at the beginning. In the screenshot you shared, it's August 28 being shown both in the August month but also in the September month since the first week of September starts on September 1st. Does it make sense? I'm not sure how I can prevent this.

@T-Dnzt Got it, no worries. Let's ignore this for now, then.

PR #248, opened 2017-08-30
Search by tags

T-Dnzt:
[Trello Task](https://trello.com/c/tOUulYpo/114-green-commons-enable-searching-by-tag-only)

# Reason for change

Currently, tags are not used in the search process. This needs to be altered in order to have tags as primary matchers.

# Changes

- Include tags in the search process
- Update the list of tags in the search filters sidebar

### DIFF
--- a/app/views/search/_filters.html.slim
+++ b/app/views/search/_filters.html.slim
@@ -25,13 +25,9 @@
         | Tags
         span.filters__clear= link_to 'Clear', root_path, class: 'gray-light pull-right'
     .tags-list
-      = link_to '#urban sustainability (245)', root_path, class: 'tags-list__tag'
-      = link_to '#solar energy (148)', root_path, class: 'tags-list__tag'
-      = link_to '#wind (136)', root_path, class: 'tags-list__tag tags-list__tag--selected'
-      = link_to '#ocean (136)', root_path, class: 'tags-list__tag tags-list__tag--selected'
-      = link_to '#solar energy (148)', root_path, class: 'tags-list__tag'
-      = link_to '#solar (148)', root_path, class: 'tags-list__tag'
-      .clearfix
+      - ActsAsTaggableOn::Tag.most_used(10).each do |tag|
+        = link_to "##{tag.name} (#{tag.taggings_count})", search_path(query: tag.name), class: 'tags-list__tag'
+        .clearfix
     .filters__tags-show-more
       = link_to 'Show more', root_path
   .filters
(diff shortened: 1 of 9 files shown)


ptrikutam:
@T-Dnzt let me know if this is too large of a request and we can create a separate PR for it and tackle it separately.

T-Dnzt:
@ptrikutam Ready for re-review. ;)

T-Dnzt:
@ptrikutam That's what I was aiming for, but I tried everything I could think of without success - ElasticSearch would simply never return what I wanted so I settled for the current version. I'll give it another try tomorrow evening :)

ptrikutam:
@T-Dnzt Not sure if this will be helpful, but maybe we could try it? https://www.elastic.co/guide/en/elasticsearch/guide/current/_finding_exact_values.html

T-Dnzt:
@ptrikutam Yeah I did try that :( The problem is I think due to the tags field being indexed as an array. When doing exact value match ("solar panels" in `tags: ["solar panels", "etc"]`) in ES, it doesn't return anything. Until I can fix that, the current implementation split "solar" and "panels" and search for those terms in the list of tags separately, and boom, it somehow works. :/

app/views/search/_filters.html.slim
```diff
@@ -25,13 +25,9 @@
...
         span.filters__clear= link_to 'Clear', root_path, class: 'gray-light pull-right'
     .tags-list
-      = link_to '#urban sustainability (245)', root_path, class: 'tags-list__tag'
-      = link_to '#solar energy (148)', root_path, class: 'tags-list__tag'
-      = link_to '#wind (136)', root_path, class: 'tags-list__tag tags-list__tag--selected'
-      = link_to '#ocean (136)', root_path, class: 'tags-list__tag tags-list__tag--selected'
-      = link_to '#solar energy (148)', root_path, class: 'tags-list__tag'
-      = link_to '#solar (148)', root_path, class: 'tags-list__tag'
-      .clearfix
+      - ActsAsTaggableOn::Tag.most_used(10).each do |tag|
```
nsanta: Can we move `ActsAsTaggableOn::Tag.most_used(10)` to the controller. 
Reason: Any complex (even simple hehe) logic should be on the view.

ptrikutam: @nsanta -- Thibault is no longer working on this project. Are you comfortable making the change you requested in a separate PR?

PR #266, opened 2017-09-11
Issue  #199 - Replace links of resources. Goes to the search page

nsanta:
* "See more" links have been replaced by "View all"  
  * Arrow icon on "View All" links have been removed
  * "View All" link now points to the search page

https://github.com/greencommons/commons/issues/199

### DIFF
--- a/app/views/shared/components/_resource_details.html.slim
+++ b/app/views/shared/components/_resource_details.html.slim
@@ -13,6 +13,5 @@
 
   .row
     .col-xs-12
-      = link_to root_path, class: 'resource-details__see-more' do
-        | See more
-        .fa.fa-arrow-right.glyphicon--left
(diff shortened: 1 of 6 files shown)


app/views/shared/components/_resource_details.html.slim
```diff
@@ -13,6 +13,5 @@
 
   .row
     .col-xs-12
-      = link_to root_path, class: 'resource-details__see-more' do
-        | See more
-        .fa.fa-arrow-right.glyphicon--left
+      = link_to search_path(query: '', filters: { model_types: {resources: 'on'}, resource_types: { resource_type.to_s.pluralize => 'on'}, sort: 'score' }), class: 'resource-details__see-more' do
```
ptrikutam: I really like the way you did this, nice one.

nsanta: I wanted to encapsulate the path builder into a helper, but it will obfuscate the code too much.

ptrikutam: All good, this works.

PR #267, opened 2017-09-11
Fix feature spec

nsanta:
@ptrikutam This PR Fixes the current broken spec.

### DIFF
--- a/spec/feature/user_searches_resources_spec.rb
+++ b/spec/feature/user_searches_resources_spec.rb
@@ -206,7 +206,12 @@ RSpec.feature 'Searching for resources', :worker, :elasticsearch do
     scenario 'users can filter by date' do
       title = Faker::Hipster.sentence
 
-      create(:resource, title: "#{title} My Resource", resource_type: :article)
+      create(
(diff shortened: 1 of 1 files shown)


spec/feature/user_searches_resources_spec.rb
```diff
@@ -206,7 +206,12 @@
     scenario 'users can filter by date' do
       title = Faker::Hipster.sentence
 
-      create(:resource, title: "#{title} My Resource", resource_type: :article)
+      create(
+        :resource,
+        title: "#{title} My Resource",
+        resource_type: :article,
+        created_at: 3.days.from_now
```
ptrikutam: Is this line necessary? The `created_at` will become the current time in the database, and below you're filtering for the latest of 3 months ago.

nsanta: It is not enough just use the current time. For some reason the resource is included in the search result. I tested it and it works with no problem. But on fails on test env. Probably it is necessary a new ticket for further research about this problem that affects only test env and just appeared a couple of days ago.

Issue #191, opened 2017-08-29
(M) Correctly reflect the number of annotations (at least, as of some date).


ptrikutam:
Requires a bit of investigation into the hypothesis API and seeing what we can get out of it.

Do we do this weekly? Nightly? 

If we do it on the front end via JS, might be able to do it dynamically.

ptrikutam:
@nsanta This one is a bit tricky, and might require some research. I'd love your help figuring this out.

Currently, we use [Hypothes.is](https://web.hypothes.is/) to link to external webpages, so that people can annotate them.

In particular, we use the Hypothes.is via link on the Resource page: 

-
 https://github.com/greencommons/commons/blob/master/app/views/resources/show.html.slim#L32-L35

- https://github.com/greencommons/commons/blob/master/app/helpers/link_helper.rb#L2-L4

It's also used in the resource card itself: https://cl.ly/1O1l0i1f0N2T

We have a place in the Resource card where we want to display the number of annotations: https://cl.ly/163i3833311Q


For the first step for this task, can you please investigate the [Hypothesis API](https://h.readthedocs.io/en/latest/) and see what is possible for making sure that "102 annotations" is replaced with a correct value on page load?

ptrikutam:
@nsanta also, we have some contacts at Hypothesis who can help us navigate the API (and may give us access to undocumented APIs) so if you're interested in speaking with them let me know.

ptrikutam:
Also, just in case it's not clear, you can see the annotation stuff in action by clicking on the "View Original" link anytime you see it: https://cl.ly/1E423s371h07

Then, at the top right of the page, you'll see a few icons https://cl.ly/3p070Z1Y1d3e

If you click the "<" drawer, it'll open all the existing annotations for a webpage. Alternatively, you can create your own by highlighting text and hitting annotate: https://cl.ly/3b0N0b3P3Z3k

PR #279, opened 2017-09-26
Issue #191 - Display total annotations of each resource in search page

nsanta:
## Reason for change

The user needs to see the real total of annotations for a resource. Currently the user sees a placeholder.

## Changes

* The resource card on search page display the total annotations
* If the resource displayed has no URL associated, the total annotation component is no rendered

## Notes
Be sure of set the `HYPOTHESIS_API_KEY` ENV var in production server instances


app/assets/javascripts/config.js.erb
```diff
@@ -0,0 +1,4 @@
+window.hypothesis = {
```
ptrikutam: I'm concerned about a top-level variable named `hypothesis` in the browser, especially if we start embedding the hypothesis JS in our page. Can we make sure this is scoped one level further? i.e.:

```javascript
window.greencommons = {
  hypothesis: {
    // ...
  }
}
```

ptrikutam: It'd be good to put future JS "config" stuff here as well.

app/assets/javascripts/components/total_annotations.jsx
```diff
@@ -0,0 +1,53 @@
+var TotalAnnotations = React.createClass({
```
ptrikutam: So we have babel built in with `react-rails`. I probably should have realized this before in your previous JSX contributions, but for what it's worth, this entire class can be rewritten like so: https://gist.github.com/ptrikutam/1c5de886d5d38b541930807bef7a8279

If you want we can walk through that in detail, or you're welcome to refactor on your own.

app/views/shared/summary_cards/_resource.html.slim
```diff
@@ -55,7 +55,7 @@
       .summary-card__more.summary-card__more--meta
         .summary-card__metadata.summary-card__metadata--last
           i.fa.fa-pencil.glyphicon--right
-          | 102 Annotations
+          = react_component 'TotalAnnotations', {url: resource.url }, class: '__inline'
```
ptrikutam: Would it make sense to include the `i.fa.fa-pencil` etc in the "Total Annotations" component? Since that icon is part of the component anyway...

app/views/shared/summary_cards/_resource.html.slim
```diff
@@ -55,7 +55,7 @@
       .summary-card__more.summary-card__more--meta
         .summary-card__metadata.summary-card__metadata--last
```
ptrikutam: Can we actually add a check here to see if there is a URL available at all? If there is no URL, there's no way for us to have any annotations, so this component wouldn't make sense to include.

spec/feature/user_searches_resources_spec.rb
```diff
@@ -13,10 +14,26 @@
...
+
+      visit new_search_path
+      within('.customer-search-form') do
+        fill_in 'query', with: title
+        find('.navbar__search-button').click
+      end
+      wait_for_ajax
+      expect(page).to have_text('Annotations')
+    end
+
```
ptrikutam: Can we also add a spec to see if it actually renders the right # of annotations? We could use VCR to record a sample URL, like so: http://www.metastatic.org/text/This%20is%20Water.pdf

That URL currently has 2 annotations. Then, we can make sure the text says "2 Annotations".

spec/rails_helper.rb
```diff
@@ -8,13 +8,28 @@
...
+  )
+
+  Capybara::Selenium::Driver.new app,
+                                 browser: :chrome,
+                                 desired_capabilities: capabilities
+end
+
 Capybara.default_driver = :webkit
-Capybara.javascript_driver = :webkit
+Capybara.javascript_driver = :headless_chrome
```
ptrikutam: For some reason this didn't actually launch a headless instance on my local machine. Not sure why, not a huge deal, but thought I'd mention it.

Gemfile
```diff
@@ -63,6 +63,8 @@ end
 
 group :test do
   gem 'capybara-webkit'
+  gem 'capybara-selenium'
+  gem 'chromedriver-helper'
```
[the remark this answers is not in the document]
ptrikutam: Come to think of it, I think this is why the specs are failing on CircleCI at the moment.

app/assets/stylesheets/pages/_default.scss
```diff
@@ -92,6 +92,10 @@
   }
 }
 
+.__inline{
+  display: inline;
+};
```
ptrikutam: Remove semicolon here.

app/assets/stylesheets/pages/_default.scss
```diff
@@ -92,6 +92,10 @@
   }
 }
 
+.__inline{
```
ptrikutam: This feels like it should belong in a different file. Perhaps a file under `app/assets/stylesheets/helpers/` titled `_utilities.scss`?

Also, perhaps you could rename the class to `inline` instead of `__inline`... Though I'm open to suggestion there.

Gemfile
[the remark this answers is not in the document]
nsanta: I'm removing this gem now. Not needed at all.

app/assets/javascripts/config.js.erb
nsanta: You are right. Great suggestion.

app/assets/stylesheets/pages/_default.scss
nsanta: I like both suggestions. I prefer name the css class as `inline`

app/assets/javascripts/components/total_annotations.jsx
nsanta: I really like your refactor. ES6 ❤️

spec/feature/user_searches_resources_spec.rb
nsanta: @ptrikutam After doing some research I found that is overcomplicated because I need to introduce 2 new dependencies (`VCR` and `puffying_billy`) and make it work together. I think that we don't need that level of assertion for now.

ptrikutam: Sounds good. Thanks for digging, and agreed -- this is not a critical assertion at this point.

app/assets/stylesheets/pages/_utilities.scss
```diff
@@ -0,0 +1,3 @@
+.inline{
```
ptrikutam: Please add a space after `inline` and before `{`

PR #284, opened 2017-09-26
Update gems

ptrikutam:
## Reason for change

Running `bummer update` helps keeps gems updated and developers sane.

## Changes

- Updated a large number of gems using `bummr`. Specifics are listed in commit messages.

## Notes

This is passing all specs, but I'm still a little nervous about some of the major version gem updates (i.e. `elasticsearch-model` and `elasticsearch-rails`, though I think it should be fine since we're on ElasticSearch 5.x on Bonsai).

Might be worth reviewing: https://github.com/elastic/elasticsearch-rails/blob/master/CHANGELOG.md

Potentially worth not merging until we run through a number of manual tests for that.


nsanta:
I will enforce perform manual tests before merge this PR into master. Also devise is a sensitive piece.

ptrikutam:
@gtourtellot Closing this one in light of our recent discussion about priorities. This isn't critical to the operation of the site and doesn't contain any new features or bug fixes.

Issue #196, opened 2017-08-29
In search results, maintain the faceting choices when doing a search on a new term, but not on a page refresh


ptrikutam:

gtourtellot:
@ptrikutam, to me this seems a good one to move to Up Next.

ptrikutam:
@gtourtellot done!

ptrikutam:
@nsanta would this require a re-architecting of the search page to accomplish? If it's going to be a larger task, we should discuss before moving forward.

nsanta:
It needs re-architecting for accomplish the second requirement. The facets selected by the user are stored as query parameters (URL).
What I can I suggest is: Have a checkbox for reset the facets selection.

ptrikutam:
Gotcha, let's avoid rearchitecting it, then. Can we store the search preferences in a cookie and have the search facets resettable with a checkbox like you suggested?

cc @gtourtellot @dweinberger @davidcmiller What do you think about the above solution?

gtourtellot:
I'd be ok with that.

dweinberger:
My preference is that the facets keep their configuration when the user
types a new phrase into the search box and does a new search (i.e., if the
person did not leave the search page and come back to it). Is that how this
would work?

That is, would the button restore the facet parameters or be a “check all”
for the facet checkboxes?

ptrikutam:
> My preference is that the facets keep their configuration when the user types a new phrase into the search box and does a new search (i.e., if the person did not leave the search page and come back to it). Is that how this would work?

That's right. However, in the scenario we're proposing, those facets would also be maintained if the person leaves the page and comes back to it (since maintaining both levels of functionality is more complex).

@dweinberger would that be OK? Or would it be better to keep functionality as is for now?

>  That is, would the button restore the facet parameters or be a “check all” for the facet checkboxes?

It'd be a "check all" to re-set all the facets to be selected.

dweinberger:
I personally think it’d be better (more useful, more defensible) to
maintain facet selections across page re-loads. So, yes to what you’re
proposing.

The “check all” button makes total sense.

PR #285, opened 2017-09-27
Issue #196 - Added search bar on search page

nsanta:
Issue: https://github.com/greencommons/commons/issues/196

## Reason for change

When the user wants to search using a new term, the filter selection is reset. 

## Changes

* The search bar on the nav bar performs new search using all filters
* The search bar on the search page will keep the filter selection


nsanta:
@ptrikutam We should disable the scan of JS(X) files on hound CI.

nsanta:
@ptrikutam Done

PR #287, opened 2017-09-28
Issue 212 resource form

nsanta:
## Reason for change

The user should be able to upload a PDF file as a resource. Also provide extra information like Author, publisher, pub date, etc.

## Changes

- The resource form displays now a tab for provide a URL or upload a PDF file.
- The file is uploaded directly to S3. The size limit for a file is 500MB.
- The new resource form let the user provide basic information: title, short content, tags and privacy setting.
- When the new resource is submitted, the app respond with the edit form page.
- The edit form provides inputs for extra information: Rights, pub date, authors, publishers, pages, ISBN.

## Notes

Before deploy to production be sure of give the value to the next ENV vars:
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY
- AWS_REGION
- AWS_UPLOADS_BUCKET

Also enable CORS on production bucket.


ptrikutam:
OK, I'm having a lot of trouble testing this, and I'm wondering if it's because of my local configuration.

I just re-generated and updated my `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Then, I set:
`AWS_REGION=us-east-1`
`AWS_UPLOADS_BUCKET` is set to the value we've discussed on Slack already.

I kept getting the following error:

https://cl.ly/0e250c3V3k2T

```
Failed to load http://<AWS_UPLOADS_BUCKET>.s3.amazonaws.com/: Response for preflight is invalid (redirect)
new:1 Uncaught (in promise) TypeError: Failed to fetch
```

I suspect it's because whenever the browser tries to do a prefetch, it gets a 307 internal redirect: https://cl.ly/0t2v0V2X2922

Then, taking the advice of [this SO post](https://stackoverflow.com/questions/35499137/fetch-api-cannot-load-the-url-response-for-preflight-is-invalid-redirect), I forced the protocol to be https in `direct_upload_input.jsx`. This seemed to give me a 200 OK from the OPTIONS request, but then I get a 403 forbidden from the POST attempt: https://cl.ly/2y1y3Q2K281y https://cl.ly/0e3T2g3Q2n2J

Here's the response I get from AWS:

```
<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
```

